PR #12, opened 2024-05-17
Add finalizer to ipam claim

maiqueb:


pkg/vmnetworkscontroller/controller.go
```diff
@@ -73,8 +80,91 @@ func (r *VirtualMachineReconciler) Reconcile(
...
+					Namespace: vmi.Namespace,
+					Name:      claimKey,
+				}
+
+				existingIPAMClaim := &ipamclaimsapi.IPAMClaim{}
+				if err := r.Client.Get(ctx, claimKey, existingIPAMClaim); err != nil {
+					if apierrors.IsNotFound(err) {
+						// we assume it had already cleaned up in the few miliseconds it took to get here ...
+						// TODO does this make sense? ... It's pretty much just for completeness.
+						continue
```
oshoval, line 115: aint we should fail here, so it will reconcile and have a new chance to create the claim ?
else if indeed it was already existing but belonging to the previous deleted VMI with same name,
and gone until we reached here as comment says, by continue we pretend all is fine, but we are missing a claim
unless there will be a reconcile because of the pod errors that will affect the VMI and then we are fine, but it is less fast and less explicit imo

oshoval: https://github.com/kubevirt/ipam-extensions/pull/42

PR #17, opened 2024-06-06
ci: ensure latest installer manifest

maiqueb:


.github/workflows/golang.yaml
```diff
@@ -25,5 +25,15 @@ jobs:
         version: v1.57.2
         args: --timeout 3m --verbose cmd/... pkg/...
 
+    - name: Ensure latest install manifest
+      run: |
+        echo "" > dist/install
+        IMG=IMG=ghcr.io/maiqueb/kubevirt-ipam-claims:latest make build-installer
```
oshoval, line 31: why double `IMG=IMG=` ?

maiqueb: whoops, because I've dropped the metaphorical ball ? 

Thanks, really good catch.

PR #21, opened 2024-06-07
deploy, kustomize: provide an OpenShift overlay

maiqueb:
Provide an overlay for OpenShift.

### DIFF
--- a/config/openshift/kustomization.yaml
+++ b/config/openshift/kustomization.yaml
@@ -1,0 +1,43 @@
+# Adds namespace to all resources.
+namespace: openshift-kubevirt-ipam-controller-system
+
+# Value of this field is prepended to the
+# names of all resources, e.g. a deployment named
+# "wordpress" becomes "alices-wordpress".
+# Note that it should also match with the prefix (text before '-') of the namespace
+# field above.
+namePrefix: kubevirt-ipam-controller-
+
+labels:
+- includeSelectors: true
+  pairs:
+    app: ipam-virt-workloads
+
+resources:
+- ../rbac
+- ../manager
+- ../webhook
+
+patches:
+- path: manager_webhook_patch.yaml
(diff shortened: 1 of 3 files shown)


config/openshift/kustomization.yaml
```diff
@@ -0,0 +1,43 @@
...
+  target:
+    kind: MutatingWebhookConfiguration
+    group: admissionregistration.k8s.io
+    version: v1
+    name: mutating-webhook-configuration
+- patch: |-
+    - op: add
+      path: /metadata/annotations
+      value:
+        service.beta.openshift.io/serving-cert-secret-name: webhook-service
```
maiqueb, line 39: need to figure out the name of the service here ... the prefix above (line 9) is not playing along here.

maybe consider removing the prefix ?...

maiqueb: Ah thanks. I had forgotten about that.

Well, it still makes sense, but, the priority surely drops quite a lot. 

It would still be useful to test drive the feature early on OpenShift though.

oshoval: changing manually only the required bit, the secret need to be "kubevirt-ipam-claims-webhook-service"
maybe you can set it as such for both U/S and D/S
without prefixing stuff (or at least not the secret) and it will work ?

(adapting to kubevirt-ipam-controller-webhook-service once time comes)

oshoval:
didnt try
but this might work and maybe we can just start by having another manifest dedicated for openshift with some flag etc

https://gist.github.com/oshoval/d5db815ebea60eb1777a838bee0530b9

oshoval:
worth to add this task to the backlog
for example so we can easily deploy potential fixes on QE clusters for quick checking if a bug is fixed
i.e istio with empty config and https://github.com/kubevirt/ipam-extensions/pull/90

maiqueb:
Superseded by #91

PR #36, opened 2024-07-04
Support primary network

qinqon:

This PR add support for ovn-kubernetes primary networks by doing a pair of things:
- Creating an IPAMClaim if there is a primary network at namespace
- Annotating the virt-launcher pod with  `k8s.ovn.org/ovn-udn-ipamclaim-reference`

TODO:
- [x] controller unit test 
- [x] webhook unit test 
- [ ] e2e test

**Release note**:
```release-note
Support ovn-kubernetes primary network feature.
```

### DIFF
--- a/pkg/config/types.go
+++ b/pkg/config/types.go
@@ -11,6 +11,8 @@ const (
 	NetworkRolePrimary NetworkRole = "primary"
 )
 
+const OVNPrimaryNetworkIPAMClaimAnnotation = "k8s.ovn.org/ovn-udn-ipamclaim-reference"
+
 type RelevantConfig struct {
 	Name               string      `json:"name"`
 	AllowPersistentIPs bool        `json:"allowPersistentIPs,omitempty"`
--- a/pkg/config/types.go
+++ b/pkg/config/types.go
@@ -5,9 +5,16 @@ import (
 	"fmt"
 )
 
+type NetworkRole string
+
+const (
+	NetworkRolePrimary NetworkRole = "primary"
+)
+
 type RelevantConfig struct {
-	Name               string `json:"name"`
-	AllowPersistentIPs bool   `json:"allowPersistentIPs,omitempty"`
+	Name               string      `json:"name"`
+	AllowPersistentIPs bool        `json:"allowPersistentIPs,omitempty"`
+	Role               NetworkRole `json:"role,omitempty"`
 }
 
 func NewConfig(nadSpec string) (*RelevantConfig, error) {
--- a/pkg/ipamclaimswebhook/podmutator.go
+++ b/pkg/ipamclaimswebhook/podmutator.go
@@ -65,6 +65,14 @@ func (a *IPAMClaimsValet) Handle(ctx context.Context, request admission.Request)
 		return admission.Errored(http.StatusBadRequest, err)
 	}
 
+	vmName, hasVMAnnotation := pod.Annotations["kubevirt.io/domain"]
+	if !hasVMAnnotation {
+		log.Info(
+			"does not have the kubevirt VM annotation",
+		)
+		return admission.Allowed("not a VM")
+	}
+
 	log.Info("webhook handling event")
 	networkSelectionElements, err := netutils.ParsePodNetworkAnnotation(pod)
 	if err != nil {
@@ -112,15 +120,6 @@ func (a *IPAMClaimsValet) Handle(ctx context.Context, request admission.Request)
 				"NAD", nadName,
 				"network", pluginConfig.Name,
 			)
-			vmName, hasVMAnnotation := pod.Annotations["kubevirt.io/domain"]
-			if !hasVMAnnotation {
-				log.Info(
-					"does not have the kubevirt VM annotation",
-					"NAD", nadName,
-					"network", pluginConfig.Name,
-				)
-				return admission.Allowed("not a VM")
-			}
 
 			vmKey := types.NamespacedName{Namespace: pod.Namespace, Name: vmName}
 			vmi := &virtv1.VirtualMachineInstance{}
--- a/pkg/ipamclaimswebhook/podmutator.go
+++ b/pkg/ipamclaimswebhook/podmutator.go
@@ -39,7 +39,9 @@ import (
 
 	virtv1 "kubevirt.io/api/core/v1"
 
+	"github.com/kubevirt/ipam-extensions/pkg/claims"
 	"github.com/kubevirt/ipam-extensions/pkg/config"
+	"github.com/kubevirt/ipam-extensions/pkg/udn"
 )
 
 // +kubebuilder:webhook:path=/mutate-v1-pod,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=ipam-claims.k8s.cni.cncf.io,admissionReviewVersions=v1,sideEffects=None
@@ -65,6 +67,8 @@ func (a *IPAMClaimsValet) Handle(ctx context.Context, request admission.Request)
 		return admission.Errored(http.StatusBadRequest, err)
 	}
 
+	log.Info("webhook handling event")
+
 	vmName, hasVMAnnotation := pod.Annotations["kubevirt.io/domain"]
 	if !hasVMAnnotation {
 		log.Info(
@@ -73,93 +77,43 @@ func (a *IPAMClaimsValet) Handle(ctx context.Context, request admission.Request)
 		return admission.Allowed("not a VM")
 	}
 
-	log.Info("webhook handling event")
 	networkSelectionElements, err := netutils.ParsePodNetworkAnnotation(pod)
 	if err != nil {
 		var goodTypeOfError *v1.NoK8sNetworkError
-		if errors.As(err, &goodTypeOfError) {
-			return admission.Allowed("no secondary networks requested")
+		if !errors.As(err, &goodTypeOfError) {
+			return admission.Errored(http.StatusBadRequest, fmt.Errorf("failed to parse pod network selection elements"))
 		}
-		return admission.Errored(http.StatusBadRequest, fmt.Errorf("failed to parse pod network selection elements"))
 	}
 
-	var (
-		hasChangedNetworkSelectionElements bool
-		podNetworkSelectionElements        = make([]v1.NetworkSelectionElement, 0, len(networkSelectionElements))
-	)
-	for _, networkSelectionElement := range networkSelectionElements {
-		nadName := types.NamespacedName{
-			Namespace: networkSelectionElement.Namespace,
-			Name:      networkSelectionElement.Name,
-		}.String()
-		log.Info(
-			"iterating network selection elements",
-			"NAD", nadName,
-		)
-		nadKey := types.NamespacedName{
-			Namespace: networkSelectionElement.Namespace,
-			Name:      networkSelectionElement.Name,
+	var newPod *corev1.Pod
+	hasChangedNetworkSelectionElements, err :=
+		ensureIPAMClaimRefAtNetworkSelectionElements(ctx, a.Client, pod, vmName, networkSelectionElements)
+	if err != nil {
+		return admission.Errored(http.StatusInternalServerError, err)
+	}
+	if hasChangedNetworkSelectionElements {
+		if newPod == nil {
+			newPod = pod.DeepCopy()
 		}
-
-		nad := v1.NetworkAttachmentDefinition{}
-		if err := a.Client.Get(context.Background(), nadKey, &nad); err != nil {
-			if k8serrors.IsNotFound(err) {
-				return admission.Allowed("NAD not found, will hang on scheduler")
-			}
+		if err := updatePodSelectionElements(newPod, networkSelectionElements); err != nil {
 			return admission.Errored(http.StatusInternalServerError, err)
 		}
-
-		pluginConfig, err := config.NewConfig(nad.Spec.Config)
-		if err != nil {
-			return admission.Errored(http.StatusInternalServerError, err)
-		}
-
-		if pluginConfig.AllowPersistentIPs {
-			log.Info(
-				"will request persistent IPs",
-				"NAD", nadName,
-				"network", pluginConfig.Name,
-			)
-
-			vmKey := types.NamespacedName{Namespace: pod.Namespace, Name: vmName}
-			vmi := &virtv1.VirtualMachineInstance{}
-			if err := a.Client.Get(context.Background(), vmKey, vmi); err != nil {
-				return admission.Errored(http.StatusInternalServerError, err)
-			}
-
-			vmiNets := vmiSecondaryNetworks(vmi)
-			networkName, foundNetworkName := vmiNets[nadKey.String()]
-			if !foundNetworkName {
-				log.Info(
-					"network name not found",
-					"NAD", nadName,
-					"network", networkName,
-				)
-				podNetworkSelectionElements = append(podNetworkSelectionElements, *networkSelectionElement)
-				continue
-			}
-
-			networkSelectionElement.IPAMClaimReference = fmt.Sprintf("%s.%s", vmName, networkName)
-			log.Info(
-				"requesting claim",
-				"NAD", nadName,
-				"network", pluginConfig.Name,
-				"claim", networkSelectionElement.IPAMClaimReference,
-			)
-			podNetworkSelectionElements = append(podNetworkSelectionElements, *networkSelectionElement)
-			hasChangedNetworkSelectionElements = true
-			continue
-		}
-		podNetworkSelectionElements = append(podNetworkSelectionElements, *networkSelectionElement)
 	}
 
-	if len(podNetworkSelectionElements) > 0 {
-		newPod, err := podWithUpdatedSelectionElements(pod, podNetworkSelectionElements)
-		if err != nil {
-			return admission.Errored(http.StatusInternalServerError, err)
+	newPrimaryNetworkIPAMClaimName, err := findNewPrimaryNetworkIPAMClaimName(ctx, a.Client, pod, vmName)
+	if err != nil {
+		return admission.Errored(http.StatusInternalServerError,
+			fmt.Errorf("failed looking for primary user defined IPAMClaim name: %v", err))
+	}
+	if newPrimaryNetworkIPAMClaimName != "" {
+		if newPod == nil {
+			newPod = pod.DeepCopy()
 		}
+		updatePodWithOVNPrimaryNetworkIPAMClaimAnnotation(newPod, newPrimaryNetworkIPAMClaimName)
+	}
 
-		if reflect.DeepEqual(newPod, pod) || !hasChangedNetworkSelectionElements {
+	if newPod != nil {
+		if reflect.DeepEqual(newPod, pod) {
 			return admission.Allowed("mutation not needed")
 		}
 
@@ -171,6 +125,7 @@ func (a *IPAMClaimsValet) Handle(ctx context.Context, request admission.Request)
 		log.Info("new pod annotations", "pod", newPod.Annotations)
 		return admission.PatchResponseFromRaw(request.Object.Raw, marshaledPod)
 	}
+
 	return admission.Allowed("carry on")
 }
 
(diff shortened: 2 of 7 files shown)


pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -65,6 +67,15 @@ func (a *IPAMClaimsValet) Handle(ctx context.Context, request admission.Request)
 		return admission.Errored(http.StatusBadRequest, err)
 	}
 
+	log.Info("webhook handling event")
```
RamLavi: Can you explain why did you move it before the "is pod VM" check on `webhook: Annotate with primary network ipamclaim` commit? 
it kinda demotes what you said you did in the  `webhook: Check first vm annotation` commit, cause now it no longer is the first thing :)

pkg/udn/primary_network.go
```diff
@@ -0,0 +1,37 @@
...
+
+	v1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kubevirt/ipam-extensions/pkg/config"
+)
+
+func FindPrimaryNetwork(ctx context.Context,
```
RamLavi, line 15: Please consider passing the nadList as an input parameter to this function. 
It would make unit-testing this function easier to do.

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -153,10 +189,17 @@ func (a *IPAMClaimsValet) Handle(ctx context.Context, request admission.Request)
...
+		if len(podNetworkSelectionElements) > 0 {
+			var err error
+			newPod, err = podWithUpdatedSelectionElements(newPod, podNetworkSelectionElements)
+			if err != nil {
+				return admission.Errored(http.StatusInternalServerError, err)
+			}
+		}
+		if primaryNetworkIpamClaimName != "" {
+			newPod.Annotations[config.OVNPrimaryNetworkIPAMClaimAnnotation] = primaryNetworkIpamClaimName
 		}
```
oshoval, line 203: nit: please consider refactoring this
maybe by doing line 194 (due to `len(podNetworkSelectionElements) > 0`) and line 201 (due to `rimaryNetworkIpamClaimName != ""`)
separately, and marking on both that a change occurred, and then 
doing the common code of line 205 if change occurred ?

pkg/vminetworkscontroller/vmi_controller.go
```diff
@@ -158,10 +159,6 @@ func (r *VirtualMachineInstanceReconciler) vmiNetworksClaimingIPAM(
 ) (map[string]string, error) {
 	vmiNets := make(map[string]string)
 	for _, net := range vmi.Spec.Networks {
-		if net.Pod != nil {
-			continue
-		}
-
 		if net.Multus != nil && !net.Multus.Default {
```
oshoval, line 162: what about primary where it is `net.Multus.Default`
maybe we dont support it yet, would we like in the future ?

oshoval (commented):
Thanks very nice
will review the UT bit later (and maybe bit more some other files)

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -65,6 +65,14 @@ func (a *IPAMClaimsValet) Handle(ctx context.Context, request admission.Request)
 		return admission.Errored(http.StatusBadRequest, err)
 	}
 
+	vmName, hasVMAnnotation := pod.Annotations["kubevirt.io/domain"]
```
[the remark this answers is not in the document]
qinqon: This webook is for pods and that is massive, so we should be sure that we don't do extra stuff for non kubevirt pods.

Perfect solution would be for mutatint webhook configuration selector to discard pods without that annotation on it.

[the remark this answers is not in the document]
qinqon: Looks like we have something for that https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-objectselector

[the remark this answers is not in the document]
qinqon: An alternative is to check `vm.kubevirt.io/name` instead (which is a label) and use 
```yaml
matchObject:
  matchExpressions:
    key: vm.kubevirt.io/name
    operator: Exists
```

pkg/udn/primary_network.go
qinqon: This can be done at not public function so we do unit test, Public intgerface should not pass nadList

pkg/udn/primary_network.go
```diff
@@ -0,0 +1,37 @@
...
+	cli client.Client,
+	namespace string) (*v1.NetworkAttachmentDefinition, error) {
+	nadList := v1.NetworkAttachmentDefinitionList{}
+	if err := cli.List(ctx, &nadList, client.InNamespace(namespace)); err != nil {
```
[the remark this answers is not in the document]
qinqon: So:
1. Controller runtime client has already an informer caching stuff behing the scenes.
2. That's interesting, but let's consider it if we hit a performance issue (Is not that we are going to have a massive number of NADs per namespace)
3. We can still have multiple NADs with that possible label, is the same scenario, so we either take the first one and not validate or validate, since there is not going to be a massive number of NADs and they are already cache at the client we can validate.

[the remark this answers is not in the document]
oshoval: > 3. We can still have multiple NADs with that possible label, is the same scenario, so we either take the first one and not validate or validate, since there is not going to be a massive number of NADs and they are already cache at the client we can validate.

What i mean is that for one NS, we can't have multiple NADs with that label, if OVN is the one that makes sure it put it on the first that is created and assert no others are created as primary on that NS (this of course involved new required logic if accepted).
Anyhow it is just food for thought in the future.
And then it gives you [2] for free, but again just nice to have as you said.

I just envision this repo to be allowed to assume stuff that is given, asserted by OVN etc.

pkg/vminetworkscontroller/vmi_controller.go
qinqon: Support that for primary UDN involve a lot of changes at different projects, so does not make sense to support it just here without sync with the others.

pkg/ipamclaimswebhook/podmutator.go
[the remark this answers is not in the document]
maiqueb: We're already doing that, aren't we ? 
https://github.com/kubevirt/ipam-extensions/pull/13/files

pkg/ipamclaimswebhook/podmutator.go
qinqon: Refactored a little so to reduce the networkselection elements copy and the condition check and make easier to follow the logic.

pkg/config/types.go
```diff
@@ -8,8 +8,11 @@ import (
 type RelevantConfig struct {
 	Name               string `json:"name"`
 	AllowPersistentIPs bool   `json:"allowPersistentIPs,omitempty"`
+	Role               string `json:"role,omitempty"`
```
maiqueb, line 16: nit: maybe make this an enum ?

pkg/udn/primary_network.go
```diff
@@ -0,0 +1,39 @@
...
+		return nil, fmt.Errorf("failed listing nads for pod namespace")
```
maiqueb, line 20: please indicate the namespace.

pkg/udn/primary_network.go
```diff
@@ -0,0 +1,39 @@
...
+	for _, nad := range nadList.Items {
+		netConfig, err := config.NewConfig(nad.Spec.Config)
+		if err != nil {
+			return nil, fmt.Errorf("failed to extract the relevant NAD information")
+		}
+		if netConfig.Role == config.NetworkRolePrimary {
+			if primaryNetworkNAD != nil {
+				return nil, fmt.Errorf("multiple primary network '%s/%s' and '%s/%s' found, "+"only one primary "+
+					"network should be configured in a given namespace",
+					nad.Namespace, nad.Name, primaryNetworkNAD.Namespace, primaryNetworkNAD.Name)
```
maiqueb, line 33: should we even care about this ? Why don't we just return the first, and let OVN-K explode in all its glory with this error ?

pkg/udn/primary_network.go
```diff
@@ -0,0 +1,39 @@
...
+			}
+			primaryNetworkNAD = ptr.To(nad)
```
maiqueb, line 35: nit:

```suggestion
			return ptr.To(nad), nil
```

pkg/udn/primary_network.go
```diff
@@ -0,0 +1,39 @@
...
+		}
+	}
+	return primaryNetworkNAD, nil
```
maiqueb, line 38: nit:

```suggestion
	return nil, nil
```

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -65,6 +65,14 @@ func (a *IPAMClaimsValet) Handle(ctx context.Context, request admission.Request)
...
+	if !hasVMAnnotation {
+		log.Info(
+			"does not have the kubevirt VM annotation",
+		)
+		return admission.Allowed("not a VM")
+	}
```
maiqueb: this shouldn't happen, right ? 

still, I'm OK to leave ... who knows ....

pkg/config/types.go
```diff
@@ -5,14 +5,17 @@ import (
...
+)
+
 type RelevantConfig struct {
 	Name               string `json:"name"`
 	AllowPersistentIPs bool   `json:"allowPersistentIPs,omitempty"`
 	Role               string `json:"role,omitempty"`
 }
 
-const NetworkRolePrimary = "primary"
-
```
maiqueb: nit: well, might as well put this in the right place in the first commit.

pkg/claims/generator.go
```diff
@@ -0,0 +1,9 @@
+package claims
+
+import (
+	"fmt"
+)
+
+func GenerateName(vmName, networkName string) string {
```
maiqueb, line 7: hm, I take "offense" at this name. At the end of the day, you're not generating anything. 

You're simply composing a name from others. 

And maybe a new file for this is a bit overkill ?... I won't insist on this second item though.

Makefile
```diff
@@ -69,7 +69,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out -v -ginkgo.v
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out -v -ginkgo.v -ginkgo.focus "${WHAT}"
```
maiqueb: this is unrelated to this commit, isn't it ?

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -73,93 +77,44 @@ func (a *IPAMClaimsValet) Handle(ctx context.Context, request admission.Request)
...
-		podNetworkSelectionElements = append(podNetworkSelectionElements, *networkSelectionElement)
 	}
 
-	if len(podNetworkSelectionElements) > 0 {
-		newPod, err := podWithUpdatedSelectionElements(pod, podNetworkSelectionElements)
-		if err != nil {
-			return admission.Errored(http.StatusInternalServerError, err)
+	newPrimaryNetworkIPAMClaimName, err := findNewPrimaryNetworkIPAMClaimName(ctx, a.Client, pod, vmName)
+	if err != nil {
+		return admission.Errored(http.StatusInternalServerError, err)
```
maiqueb, line 107: maybe compose the error ?

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -73,93 +77,44 @@ func (a *IPAMClaimsValet) Handle(ctx context.Context, request admission.Request)
...
+	}
+	if newPrimaryNetworkIPAMClaimName != "" {
+		if newPod == nil {
+			newPod = pod.DeepCopy()
 		}
+		newPod.Annotations[config.OVNPrimaryNetworkIPAMClaimAnnotation] = newPrimaryNetworkIPAMClaimName
+	}
```
maiqueb, line 114: IMHO we should move this to an helper.

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -195,7 +151,17 @@ func vmiSecondaryNetworks(vmi *virtv1.VirtualMachineInstance) map[string]string
...
-func podWithUpdatedSelectionElements(pod *corev1.Pod, networks []v1.NetworkSelectionElement) (*corev1.Pod, error) {
+// returns the name of the kubevirt VM pod network
+func vmiPodNetworkName(vmi *virtv1.VirtualMachineInstance) string {
+	for _, network := range vmi.Spec.Networks {
+		if network.Pod != nil {
+			return network.Name
+		}
+	}
+	return ""
+}
```
maiqueb, line 162: please move this to the end of the file, to minimize the changes to the code.

pkg/vminetworkscontroller/vmi_controller.go
oshoval: should a release note be added?

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -204,3 +170,117 @@ func podWithUpdatedSelectionElements(pod *corev1.Pod, networks []v1.NetworkSelec
...
+func ensureIPAMClaimRefAtNetworkSelectionElements(ctx context.Context,
+	cli client.Client, pod *corev1.Pod, vmName string,
+	networkSelectionElements []*v1.NetworkSelectionElement) (changed bool, response string, err error) {
+	log := logf.FromContext(ctx)
+	hasChangedNetworkSelectionElements := false
+	for i, networkSelectionElement := range networkSelectionElements {
+		nadName := types.NamespacedName{
+			Namespace: networkSelectionElement.Namespace,
+			Name:      networkSelectionElement.Name,
+		}.String()
```
maiqueb, line 183: nit: wouldn't it be simpler to %s.%s ? Don't we have an helper for that ? 

at least compose the `nadKey` before, and here you can just do `nadKey.String()`

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -204,3 +170,117 @@ func podWithUpdatedSelectionElements(pod *corev1.Pod, networks []v1.NetworkSelec
 	newPod.Annotations[v1.NetworkAttachmentAnnot] = string(newNets)
 	return newPod, nil
 }
+
+func ensureIPAMClaimRefAtNetworkSelectionElements(ctx context.Context,
```
maiqueb, line 174: I really dislike the fact this returns 3 things.

And the fact we use this `response` parameter just to indicate something that IMO looks like an error is very puzzling to me.

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -204,3 +170,117 @@ func podWithUpdatedSelectionElements(pod *corev1.Pod, networks []v1.NetworkSelec
...
+	if pod.Annotations[config.OVNPrimaryNetworkIPAMClaimAnnotation] != "" {
+		return "", nil
+	}
+	primaryNetworkNAD, err := udn.FindPrimaryNetwork(ctx, cli, pod.Namespace)
+	if err != nil {
+		return "", err
+	}
+	if primaryNetworkNAD == nil {
+		return "", nil
+	}
```
maiqueb, line 258: nit: maybe coalesce these ? 

```suggestion
        if err != nil || primaryNetworkNAD == nil {
		return "", err
	}
```

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -204,3 +170,117 @@ func podWithUpdatedSelectionElements(pod *corev1.Pod, networks []v1.NetworkSelec
...
+	)
+	vmKey := types.NamespacedName{Namespace: pod.Namespace, Name: vmName}
+	vmi := &virtv1.VirtualMachineInstance{}
+	if err := cli.Get(context.Background(), vmKey, vmi); err != nil {
+		return "", err
+	}
+
+	networkName := vmiPodNetworkName(vmi)
+	if networkName == "" {
+		log.Info("vmi has no pod network primary UDN ipam claim will not be created")
```
maiqueb, line 281: please print the namespaced name of the VMI that doesn't have a pod network name.

pkg/ipamclaimswebhook/podmutator_test.go
```diff
@@ -111,25 +111,80 @@ var _ = Describe("KubeVirt IPAM launcher pod mutato machine", Serial, func() {
...
+			inputNADs: []*nadv1.NetworkAttachmentDefinition{
+				dummyNAD(nadName),
+			},
 			inputPod: &corev1.Pod{},
 			expectedAdmissionResponse: admission.Response{
 				AdmissionResponse: admissionv1.AdmissionResponse{
 					Allowed: true,
 					Result: &metav1.Status{
-						Message: "no secondary networks requested",
+						Message: "not a VM",
```
maiqueb, line 125: what ? I just don't understand how a test about "pod not requesting secondary attachments and no primary user defined network is accepted" can expect to have a "not a VM" result.

The description should match the expectation - i.e. "non VM pod ..."

pkg/ipamclaimswebhook/podmutator.go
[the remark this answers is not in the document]
RamLavi: So am I correct in understanding that this check is unnecessary?

Makefile
```diff
@@ -69,7 +69,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out -v -ginkgo.v -ginkgo.focus "${WHAT}"
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out -v -ginkgo.v -ginkgo.focus="${WHAT}"
```
maiqueb, line 72: might as well make this right in the first place, rather than adding it wrong, then correcting it.

pkg/vminetworkscontroller/vmi_controller.go
```diff
@@ -193,8 +190,27 @@ func (r *VirtualMachineInstanceReconciler) vmiNetworksClaimingIPAM(
 			if nadConfig.AllowPersistentIPs {
 				vmiNets[net.Name] = nadConfig.Name
 			}
+		} else if net.Pod != nil {
```
maiqueb, line 193: can we turn this around ? i.e. handle the pod network **first** ?

maiqueb (changes requested):
I have left a lot of comments, but most of them are nits. You can choose to ignore most of them.

I really dislike the signature of one function you've introduced though. And I have to say I don't follow that part of the code well because of that function.

oshoval:
is it fine to mention https://github.com/openshift/enhancements/pull/1673 on PR desc ?

pkg/vminetworkscontroller/vmi_controller.go
oshoval: consider please adding the new logic under this condition in a function

pkg/vminetworkscontroller/vmi_controller.go
oshoval: @maiqueb 
you might want to add note about it on https://github.com/openshift/enhancements/pull/1673 please ?

pkg/udn/primary_network.go
qinqon: Let's do that.

pkg/ipamclaimswebhook/podmutator.go
qinqon: > this shouldn't happen, right ?
> 

Yep, but...

> still, I'm OK to leave ... who knows ....

pkg/claims/generator.go
qinqon: I will rename it to `ComposeName` and move it to claims.go

Makefile
qinqon: Right will remove it.

pkg/ipamclaimswebhook/podmutator.go
maiqueb: let's keep it... 

We can drop it later.

pkg/vminetworkscontroller/vmi_controller.go
maiqueb: nah good point; if you can turn this into an helper, it would for sure be better. 

It is not a must from my perspective, can be tackled as a follow-up IMHO.

pkg/ipamclaimswebhook/podmutator.go
qinqon: I don't like it either, but I was trying to maintain the original behaviour where is returning `NAD not found, will hang on scheduler` if there is no nad for the NSE, we can just return `nil, nil` but logging it and the webhook response will be "carry on"

pkg/ipamclaimswebhook/podmutator.go
qinqon: Let's not make things smarter, and keep things separated.

pkg/ipamclaimswebhook/podmutator_test.go
qinqon: Adding to the description `pod not beloging to a VM` like other tests.

Makefile
qinqon: I have just remove it.

pkg/udn/primary_network.go
RamLavi: What I mean is, Handle(webhook)/Reconcile(Controller) can list the nads, mapping them once:
```
map[string]*NAD{
   'ns/nadName1`: *NAD
   'ns/nadName2`: *NAD
   `primary`: *NAD
```
then just send the map to each function case (secondary, primary)..
But only if it works for you. If it doesn't add up them I can try it on a follow up PR..

pkg/ipamclaimswebhook/podmutator.go
RamLavi: I suggest you return a new error type like "NoNadNotOurProblem" (I'm bad with names but you get my drift) and expect that case outside of `ensureIPAMClaimRefAtNetworkSelectionElements`, returning the AdmissionAllowed

pkg/vminetworkscontroller/vmi_controller.go
qinqon: I will do a pair of functions to make is symmetric, one for for secondary and other for primary UDNs

pkg/ipamclaimswebhook/podmutator.go
qinqon: > I suggest you return a new error type like "NoNadIgnoreReqest" and expect that case outside of `ensureIPAMClaimRefAtNetworkSelectionElements`, returning the AdmissionAllowed
> 
> ```
> var ErrNoNadIgnoreRequest = errors.New("NAD not found, will hang on scheduler")
> ```
> 
> on hasChangedNetworkSelectionElements function:
> 
> ```
> if k8serrors.IsNotFound(err) {
>   return ErrNoNadIgnoreRequest
> }
> ```
> 
> then outside of it:
> 
> ```
> hasChangedNetworkSelectionElements, err :=
>     ensureIPAMClaimRefAtNetworkSelectionElements(ctx, a.Client, pod, vmName, networkSelectionElements)
> if err != nil {
>     if errors.Is(err, ErrNoNadIgnoreRequest) {
>         return admission.Allowed(ErrNoNadIgnoreRequest.Error())
>     }
>     return admission.Errored(http.StatusInternalServerError, err)
> }
> ```

Puff I think this is overkill just to return something at the webhook success, I want to know if returning that specific string is really necessary or logging it is enough.

maiqueb: hm, returning bool, error also makes me frown...

qinqon: ok, simplified since we don't care about th emssage at webhook.Allowed

pkg/config/types.go
```diff
@@ -5,9 +5,19 @@ import (
 	"fmt"
 )
 
+type NetworkRole string
+
+const (
+	NetworkRolePrimary   NetworkRole = "primary"
+	NetworkRoleSecondary NetworkRole = "secondary"
```
maiqueb, line 12: not used

pkg/ipamclaimswebhook/podmutator.go
maiqueb: no sure you saw it and chose to ignore it, or just didn't see it. 

> hm, returning bool, error also makes me frown...

Can we try to improve this API ?

qinqon: What the issue with that ? is a function that try to update a map and return if the map was updated or not.

maiqueb: you can do the exact same thing with just error, or returning the map and letting the client understand if it changed or not. 

if your function does "update a map and return if the map was updated or not" your function is doing 2 things. The function should do one. Update the map. 

We can improve it later on if this is the only item under question.

qinqon: openned an issue to simplify things later on https://github.com/kubevirt/ipam-extensions/issues/63

RamLavi:
I like how this makes the webhook Handle function much more readable 
/lgtm

oshoval:
i guess we want it to be deployed by CNAO
for that need to create a tag so it would be bumped on CNAO

PR #40, opened 2024-07-08
Fix Finalizer logic

oshoval:
Remove the finalizer from the ipam claim on the following cases:

VMI events:
1. VM that is marked for deletion, and its VMI is gone.
2. Standalone VMI, that is marked for deletion, and it's virt-launcher is gone.
3. Standalone VMI that was deleted.

VM events:
1. VM is deleted (happens on background delete).
2. VM is marked for deletion, and the VMI is deleted (foreground delete).

(might remind VMI events, but it can happen for stopped VM so we should handle
it as well as VM event).

Changes beside that:
1. Add missing `BlockOwnerDeletion` that affects foreground delete.
Owner which is either VM or VMI, will be deleted only after the
owned ipam claim is deleted.
We remove the finalizer from the ipam claim, once the claim is ready to be deleted.
2. Add missing `Controller` field, since the claims are created by this controller.
3. Fix `IPAMClaimsFromNamespace` to list claims only per given namespace.

Fixes #

A follow-up PR that upon controller restart, we look for dangling claims
and remove their finalizer. For example if a VM was deleted while controller was down.
Controller will restart, and won't get event on the VM that was deleted, because it gets info about the current
cluster state. Therefore we should clean finalizer from those orphan ipam claims.
PR: https://github.com/kubevirt/ipam-extensions/pull/48

**Checklist**

This checklist is not enforcing, but it's a reminder of items that could be relevant to every PR.
Approvers are expected to review this list.

- [ ] Design: A [design document](https://github.com/kubevirt/community/tree/main/design-proposals) was considered and is present (link) or not required
- [ ] PR: The PR description is expressive enough and will help future contributors
- [ ] Code: [Write code that humans can understand](https://en.wikiquote.org/wiki/Martin_Fowler#code-for-humans) and [Keep it simple](https://en.wikipedia.org/wiki/KISS_principle)
- [ ] Refactor: You have [left the code cleaner than you found it (Boy Scout Rule)](https://learning.oreilly.com/library/view/97-things-every/9780596809515/ch08.html)
- [ ] Upgrade: Impact of this change on upgrade flows was considered and addressed if required
- [ ] Testing: New code requires [new unit tests](https://github.com/kubevirt/kubevirt/blob/main/docs/reviewer-guide.md#when-is-a-pr-good-enough). New features and bug fixes require at least on e2e test
- [ ] Documentation: A [user-guide update](https://github.com/kubevirt/user-guide/) was considered and is present (link) or not required. You want a user-guide update if it's a user facing feature / API change.
- [ ] Community: Announcement to [kubevirt-dev](https://groups.google.com/g/kubevirt-dev/) was considered

**Release note**:

```release-note

```

### DIFF
--- a/pkg/claims/claims.go
+++ b/pkg/claims/claims.go
@@ -1,0 +1,43 @@
+package claims
+
+import (
+	"context"
+	"fmt"
+
+	apitypes "k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	ipamclaimsapi "github.com/k8snetworkplumbingwg/ipamclaims/pkg/crd/ipamclaims/v1alpha1"
+
+	virtv1 "kubevirt.io/api/core/v1"
+)
+
+const KubevirtVMFinalizer = "kubevirt.io/persistent-ipam"
+
+func Cleanup(c client.Client, vmiKey apitypes.NamespacedName) error {
+	ipamClaims := &ipamclaimsapi.IPAMClaimList{}
+	listOpts := []client.ListOption{
+		client.InNamespace(vmiKey.Namespace),
+		OwnedByVMLabel(vmiKey.Name),
+	}
+	if err := c.List(context.Background(), ipamClaims, listOpts...); err != nil {
+		return fmt.Errorf("could not get list of IPAMClaims owned by VM %q: %w", vmiKey.String(), err)
+	}
+
+	for _, claim := range ipamClaims.Items {
+		if controllerutil.RemoveFinalizer(&claim, KubevirtVMFinalizer) {
+			if err := c.Update(context.Background(), &claim, &client.UpdateOptions{}); err != nil {
+				return client.IgnoreNotFound(err)
+			}
+		}
+	}
+	return nil
+}
+
+func OwnedByVMLabel(vmiName string) client.MatchingLabels {
+	return map[string]string{
+		virtv1.VirtualMachineLabel: vmiName,
+	}
+}
--- a/pkg/vminetworkscontroller/vmi_controller.go
+++ b/pkg/vminetworkscontroller/vmi_controller.go
@@ -17,7 +17,6 @@ import (
 
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
@@ -26,10 +25,9 @@ import (
 
 	virtv1 "kubevirt.io/api/core/v1"
 
+	"github.com/kubevirt/ipam-extensions/pkg/claims"
 	"github.com/kubevirt/ipam-extensions/pkg/config"
 )
-
-const kubevirtVMFinalizer = "kubevirt.io/persistent-ipam"
 
 // VirtualMachineInstanceReconciler reconciles a VirtualMachineInstance object
 type VirtualMachineInstanceReconciler struct {
@@ -69,7 +67,7 @@ func (r *VirtualMachineInstanceReconciler) Reconcile(
 	}
 
 	if shouldCleanFinalizers(vmi, vm) {
-		if err := r.Cleanup(request.NamespacedName); err != nil {
+		if err := claims.Cleanup(r.Client, request.NamespacedName); err != nil {
 			return controllerruntime.Result{}, fmt.Errorf("failed removing the IPAMClaims finalizer: %w", err)
 		}
 		return controllerruntime.Result{}, nil
@@ -92,8 +90,8 @@ func (r *VirtualMachineInstanceReconciler) Reconcile(
 				Name:            claimKey,
 				Namespace:       vmi.Namespace,
 				OwnerReferences: []metav1.OwnerReference{ownerInfo},
-				Finalizers:      []string{kubevirtVMFinalizer},
-				Labels:          ownedByVMLabel(vmi.Name),
+				Finalizers:      []string{claims.KubevirtVMFinalizer},
+				Labels:          claims.OwnedByVMLabel(vmi.Name),
 			},
 			Spec: ipamclaimsapi.IPAMClaimSpec{
 				Network: netConfigName,
@@ -198,33 +196,6 @@ func (r *VirtualMachineInstanceReconciler) vmiNetworksClaimingIPAM(
 		}
 	}
 	return vmiNets, nil
-}
-
-func (r *VirtualMachineInstanceReconciler) Cleanup(vmiKey apitypes.NamespacedName) error {
-	ipamClaims := &ipamclaimsapi.IPAMClaimList{}
-	listOpts := []client.ListOption{
-		client.InNamespace(vmiKey.Namespace),
-		ownedByVMLabel(vmiKey.Name),
-	}
-	if err := r.Client.List(context.Background(), ipamClaims, listOpts...); err != nil {
-		return fmt.Errorf("could not get list of IPAMClaims owned by VM %q: %w", vmiKey.String(), err)
-	}
-
-	for _, claim := range ipamClaims.Items {
-		removedFinalizer := controllerutil.RemoveFinalizer(&claim, kubevirtVMFinalizer)
-		if removedFinalizer {
-			if err := r.Client.Update(context.Background(), &claim, &client.UpdateOptions{}); err != nil {
-				return client.IgnoreNotFound(err)
-			}
-		}
-	}
-	return nil
-}
-
-func ownedByVMLabel(vmiName string) client.MatchingLabels {
-	return map[string]string{
-		virtv1.VirtualMachineLabel: vmiName,
-	}
 }
 
 func shouldCleanFinalizers(vmi *virtv1.VirtualMachineInstance, vm *virtv1.VirtualMachine) bool {
--- a/pkg/vminetworkscontroller/vmi_controller.go
+++ b/pkg/vminetworkscontroller/vmi_controller.go
@@ -8,6 +8,8 @@ import (
 
 	"github.com/go-logr/logr"
 
+	"k8s.io/utils/ptr"
+
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -56,30 +58,25 @@ func (r *VirtualMachineInstanceReconciler) Reconcile(
 	defer cancel()
 	err := r.Client.Get(contextWithTimeout, request.NamespacedName, vmi)
 	if apierrors.IsNotFound(err) {
-		r.Log.Info("could not retrieve VMI - will cleanup its IPAMClaims")
-		if err := r.Cleanup(request.NamespacedName); err != nil {
-			return controllerruntime.Result{}, fmt.Errorf("error removing the IPAMClaims finalizer: %w", err)
-		}
-		return controllerruntime.Result{}, nil
+		vmi = nil
 	} else if err != nil {
 		return controllerruntime.Result{}, err
 	}
 
-	var ownerInfo metav1.OwnerReference
-	vm := &virtv1.VirtualMachine{}
-	contextWithTimeout, cancel = context.WithTimeout(ctx, time.Second)
-	defer cancel()
-	if err := r.Client.Get(contextWithTimeout, request.NamespacedName, vm); apierrors.IsNotFound(err) {
-		r.Log.Info("Corresponding VM not found", "vm", request.NamespacedName)
-		ownerInfo = metav1.OwnerReference{APIVersion: vmi.APIVersion, Kind: vmi.Kind, Name: vmi.Name, UID: vmi.UID}
-	} else if err == nil {
-		ownerInfo = metav1.OwnerReference{APIVersion: vm.APIVersion, Kind: vm.Kind, Name: vm.Name, UID: vm.UID}
-	} else {
-		return controllerruntime.Result{}, fmt.Errorf(
-			"error to get VMI %q corresponding VM: %w",
-			request.NamespacedName,
-			err,
-		)
+	vm, err := getOwningVM(ctx, r.Client, request.NamespacedName)
+	if err != nil {
+		return controllerruntime.Result{}, err
+	}
+
+	if shouldCleanFinalizers(vmi, vm) {
+		if err := r.Cleanup(request.NamespacedName); err != nil {
+			return controllerruntime.Result{}, fmt.Errorf("failed removing the IPAMClaims finalizer: %w", err)
+		}
+		return controllerruntime.Result{}, nil
+	}
+
+	if vmi == nil {
+		return controllerruntime.Result{}, nil
 	}
 
 	vmiNetworks, err := r.vmiNetworksClaimingIPAM(ctx, vmi)
@@ -87,6 +84,7 @@ func (r *VirtualMachineInstanceReconciler) Reconcile(
 		return controllerruntime.Result{}, err
 	}
 
+	ownerInfo := ownerReferenceFor(vmi, vm)
 	for logicalNetworkName, netConfigName := range vmiNetworks {
 		claimKey := fmt.Sprintf("%s.%s", vmi.Name, logicalNetworkName)
 		ipamClaim := &ipamclaimsapi.IPAMClaim{
@@ -148,7 +146,7 @@ func onVMIPredicates() predicate.Funcs {
 			return true
 		},
 		UpdateFunc: func(updateEvent event.UpdateEvent) bool {
-			return false
+			return true
 		},
 		GenericFunc: func(event.GenericEvent) bool {
 			return false
@@ -226,5 +224,50 @@ func (r *VirtualMachineInstanceReconciler) Cleanup(vmiKey apitypes.NamespacedNam
 func ownedByVMLabel(vmiName string) client.MatchingLabels {
 	return map[string]string{
 		virtv1.VirtualMachineLabel: vmiName,
+	}
+}
+
+func shouldCleanFinalizers(vmi *virtv1.VirtualMachineInstance, vm *virtv1.VirtualMachine) bool {
+	if vm != nil {
+		// VMI is gone and VM is marked for deletion
+		return vmi == nil && vm.DeletionTimestamp != nil
+	} else {
+		// VMI is gone or VMI is marked for deletion, virt-launcher is gone
+		return vmi == nil || (vmi.DeletionTimestamp != nil && len(vmi.Status.ActivePods) == 0)
+	}
+}
+
+func ownerReferenceFor(vmi *virtv1.VirtualMachineInstance, vm *virtv1.VirtualMachine) metav1.OwnerReference {
+	var obj client.Object
+	if vm != nil {
+		obj = vm
+	} else {
+		obj = vmi
+	}
+
+	aPIVersion := obj.GetObjectKind().GroupVersionKind().Group + "/" + obj.GetObjectKind().GroupVersionKind().Version
+	return metav1.OwnerReference{
+		APIVersion:         aPIVersion,
+		Kind:               obj.GetObjectKind().GroupVersionKind().Kind,
+		Name:               obj.GetName(),
+		UID:                obj.GetUID(),
+		Controller:         ptr.To(true),
+		BlockOwnerDeletion: ptr.To(true),
+	}
+}
+
+// Gets the owning VM if any. for simplicity it just try to fetch the VM,
+// even when the VMI exists, instead of parsing ownerReferences and handling differently the nil VMI case.
+func getOwningVM(ctx context.Context, c client.Client, name apitypes.NamespacedName) (*virtv1.VirtualMachine, error) {
+	contextWithTimeout, cancel := context.WithTimeout(ctx, time.Second)
+	defer cancel()
+
+	vm := &virtv1.VirtualMachine{}
+	if err := c.Get(contextWithTimeout, name, vm); err == nil {
+		return vm, nil
+	} else if apierrors.IsNotFound(err) {
+		return nil, nil
+	} else {
+		return nil, fmt.Errorf("failed getting VM %q: %w", name, err)
 	}
 }
--- a/pkg/vminetworkscontroller/vmi_controller_test.go
+++ b/pkg/vminetworkscontroller/vmi_controller_test.go
@@ -29,6 +29,8 @@ import (
 
 	ipamclaimsapi "github.com/k8snetworkplumbingwg/ipamclaims/pkg/crd/ipamclaims/v1alpha1"
 	nadv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+
+	"github.com/kubevirt/ipam-extensions/pkg/claims"
 )
 
 func TestController(t *testing.T) {
@@ -134,7 +136,7 @@ var _ = Describe("VMI IPAM controller", Serial, func() {
 		if len(config.expectedIPAMClaims) > 0 {
 			ipamClaimList := &ipamclaimsapi.IPAMClaimList{}
 
-			Expect(mgr.GetClient().List(context.Background(), ipamClaimList, ownedByVMLabel(vmName))).To(Succeed())
+			Expect(mgr.GetClient().List(context.Background(), ipamClaimList, claims.OwnedByVMLabel(vmName))).To(Succeed())
 			Expect(ipamClaimsCleaner(ipamClaimList.Items...)).To(ConsistOf(config.expectedIPAMClaims))
 		}
 	},
@@ -148,8 +150,8 @@ var _ = Describe("VMI IPAM controller", Serial, func() {
 					ObjectMeta: metav1.ObjectMeta{
 						Name:       fmt.Sprintf("%s.%s", vmName, "randomnet"),
 						Namespace:  namespace,
-						Finalizers: []string{kubevirtVMFinalizer},
-						Labels:     ownedByVMLabel(vmName),
+						Finalizers: []string{claims.KubevirtVMFinalizer},
+						Labels:     claims.OwnedByVMLabel(vmName),
 						OwnerReferences: []metav1.OwnerReference{{
 							APIVersion:         "kubevirt.io/v1",
 							Kind:               "VirtualMachine",
@@ -201,8 +203,8 @@ var _ = Describe("VMI IPAM controller", Serial, func() {
 				ObjectMeta: metav1.ObjectMeta{
 					Name:       fmt.Sprintf("%s.%s", vmName, "randomnet"),
 					Namespace:  namespace,
-					Finalizers: []string{kubevirtVMFinalizer},
-					Labels:     ownedByVMLabel(vmName),
+					Finalizers: []string{claims.KubevirtVMFinalizer},
+					Labels:     claims.OwnedByVMLabel(vmName),
 				},
 				Spec: ipamclaimsapi.IPAMClaimSpec{Network: "doesitmatter?"},
 			},
@@ -211,7 +213,7 @@ var _ = Describe("VMI IPAM controller", Serial, func() {
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      "vm1.randomnet",
 						Namespace: "ns1",
-						Labels:    ownedByVMLabel(vmName),
+						Labels:    claims.OwnedByVMLabel(vmName),
 					},
 					Spec: ipamclaimsapi.IPAMClaimSpec{Network: "doesitmatter?"},
 				},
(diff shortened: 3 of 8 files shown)


pkg/vmnetworkscontroller/vm_controller.go
```diff
@@ -0,0 +1,67 @@
...
+		},
+		DeleteFunc: func(event.DeleteEvent) bool {
+			return true
+		},
+		UpdateFunc: func(updateEvent event.UpdateEvent) bool {
+			return false
+		},
+		GenericFunc: func(event.GenericEvent) bool {
+			return false
+		},
```
oshoval, line 65: is it fine those two are false ?
i copy pasted it from VMI controiller, and we also need update events arent we?
if so it means we need to change it there as well

maiqueb:
> Note: since we deploy the merged main/HEAD version before cluster-sync, and not waiting with the e2e tests until the version is stabilized, the first test that spin VM might fail depend on the changes between the PR and main branch
> 
> waiting for the component to finish deploying is enough to fix it [82466a8#diff-e3928d4239da3d4a2cac54301cc59c71dcdde5ed062dde916f3ccd1359e64864R78](https://github.com/kubevirt/ipam-extensions/commit/82466a82a9a744fab1ebdfaa8295c429aef218e3#diff-e3928d4239da3d4a2cac54301cc59c71dcdde5ed062dde916f3ccd1359e64864R78)
> 
> but imo we should also not deploy the main/HEAD and to later take into account that VM should be robust to be created during a version upgrade (unless in this case we can just ask to recreate the VM, which works, because not every upgrade can be robust to creation during upgrade) it is also not easy to simulate it

Push a separate PR for this, will you ?

oshoval:
> > Note: since we deploy the merged main/HEAD version before cluster-sync, and not waiting with the e2e tests until the version is stabilized, the first test that spin VM might fail depend on the changes between the PR and main branch
> > waiting for the component to finish deploying is enough to fix it [82466a8#diff-e3928d4239da3d4a2cac54301cc59c71dcdde5ed062dde916f3ccd1359e64864R78](https://github.com/kubevirt/ipam-extensions/commit/82466a82a9a744fab1ebdfaa8295c429aef218e3#diff-e3928d4239da3d4a2cac54301cc59c71dcdde5ed062dde916f3ccd1359e64864R78)
> > but imo we should also not deploy the main/HEAD and to later take into account that VM should be robust to be created during a version upgrade (unless in this case we can just ask to recreate the VM, which works, because not every upgrade can be robust to creation during upgrade) it is also not easy to simulate it
> 
> Push a separate PR for this, will you ?

ye for sure, planned to

Btw maybe the "sync" func that should handle the controller restart, wrt removing finalizers and so on,
can also fix problems like this, by aligning the current state with the desired state

oshoval:
> Push a separate PR for this, will you ?

https://github.com/kubevirt/ipam-extensions/pull/45

oshoval:
Arranged commits, and updated PR description, going to add foreground deletion tests (didn't test manually yet all flows)
Basically the essence is ready for review (even to just see we agree on the direction please for now)
Thanks

oshoval:
Basically it is ready
We can still add more unit tests for the VMI controller finalizer removal flows upon needs (e2e does test them already),
consolidate the helper functions that repeat on the unit tests, i will create a file named 
`/pkg/testobjects/testobjects.go` is it ok ?
and to do a follow-up for dangling claims upon restart, as mentioned on PR desc.

oshoval:
going to rebase, so please wait a moment with review

oshoval:
rebased

maiqueb (changes requested):
Thanks @oshoval . Really good PR.

I only have some minor nits.

I also want to ensure we properly test the finalizer / foreground deletion; while your work has been amazing, I do feel we are missing ensuring the IPAM claims are not deleted until the actual VMI is gone (when we have a VM).

What do you think of this test:
- create VM (which spins up a VMI)
- add a user based finalizer to the VM (or have the VMI created with the finalizer)
- delete the VM (if needed, remember to remove its finalizer)
  - ensure the claims are **not** deleted
- delete the VMI's finalizer
  - ensure the claims are deleted after the VMI is deleted

pkg/vminetworkscontroller/vmi_controller_test.go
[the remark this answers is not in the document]
oshoval: can't squash easily because there are some commits in the middle that did refactor, and the later commit depends on them
if you dont insist please i prefer to keep as is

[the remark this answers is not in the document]
maiqueb: I guess all good.

oshoval:
> I also want to ensure we properly test the finalizer / foreground deletion; while your work has been amazing, I do feel we are missing ensuring the IPAM claims are not deleted until the actual VMI is gone (when we have a VM).
> 
> What do you think of this test:
> 
> * create VM (which spins up a VMI)
> * add a user based finalizer to the VM (or have the VMI created with the finalizer)
> * delete the VM (if needed, remember to remove its finalizer)
>   
>   * ensure the claims are **not** deleted
> * delete the VMI's finalizer
>   
>   * ensure the claims are deleted after the VMI is deleted

IIUC you mean background delete of a VM, where there is a VMI, and you want to make sure that the ipam claim is not deleted until the VMI is gone ?
If so, it is given, because the VM waits to the VMI to be deleted, enforced by the kubevirt finalizers logic,
and since it is background delete, the ipam claim will be deleted only after the VM,
hence after the VMI.
I think we don't need to over test as such, unless you have other cases or still think otherwise,
what we can consider is to add "len (claims) != 0" in some places, for example when we stop a VM, before we delete it.
Wdyt?

maiqueb:
> > I also want to ensure we properly test the finalizer / foreground deletion; while your work has been amazing, I do feel we are missing ensuring the IPAM claims are not deleted until the actual VMI is gone (when we have a VM).
> > What do you think of this test:
> > 
> > * create VM (which spins up a VMI)
> > * add a user based finalizer to the VM (or have the VMI created with the finalizer)
> > * delete the VM (if needed, remember to remove its finalizer)
> >   
> >   * ensure the claims are **not** deleted
> > * delete the VMI's finalizer
> >   
> >   * ensure the claims are deleted after the VMI is deleted
> 
> IIUC you mean background delete of a VM, where there is a VMI, and you want to make sure that the ipam claim is not deleted until the VMI is gone ? If so, it is given, because the VM waits to the VMI to be deleted, enforced by the kubevirt finalizers logic, and since it is background delete, the ipam claim will be deleted only after the VM, hence after the VMI. I think we don't need to over test as such, unless you have other cases or still think otherwise, what we can consider is to add "len (claims) != 0" in some places (with `Consistently`), for example when we stop a VM, before we delete it. Wdyt?

No, I mean foreground delete of a VM. I want to be able to say I have a test that ensures you do not release the IPs on the claims to the pool before the workload is really gone.

And I can't rely on mechanisms outside this repo. We need a test here.

oshoval:
> > IIUC you mean background delete of a VM, where there is a VMI, and you want to make sure that the ipam claim is not deleted until the VMI is gone ? If so, it is given, because the VM waits to the VMI to be deleted, enforced by the kubevirt finalizers logic, and since it is background delete, the ipam claim will be deleted only after the VM, hence after the VMI. I think we don't need to over test as such, unless you have other cases or still think otherwise, what we can consider is to add "len (claims) != 0" in some places (with `Consistently`), for example when we stop a VM, before we delete it. Wdyt?
> 
> No, I mean foreground delete of a VM. I want to be able to say I have a test that ensures you do not release the IPs on the claims to the pool before the workload is really gone.
> 
> And I can't rely on mechanisms outside this repo. We need a test here.

I understand, imho it worth to check it once manually, feels to me really overkill to add such test, that runs each and everytime, please let me know which direction you prefer and i will do it.

maiqueb:
> > No, I mean foreground delete of a VM. I want to be able to say I have a test that ensures you do not release the IPs on the claims to the pool before the workload is really gone.
> > And I can't rely on mechanisms outside this repo. We need a test here.
> 
> I understand, imho it worth to check it once manually, feels to me overkill to add such test, that runs each and everytime, but i will add it. Basically i think that we can take kubevirt as "iron" (assume it works correct, as we are doing usually) but this repo might have bugs, so due to this reason it might be good to have such test, it is just a bit deeper test.

Please add the test.

We are extending KubeVirt's behavior, and we need to have an e2e test that shows the integration works as expected.

qinqon:
> > > No, I mean foreground delete of a VM. I want to be able to say I have a test that ensures you do not release the IPs on the claims to the pool before the workload is really gone.
> > > And I can't rely on mechanisms outside this repo. We need a test here.
> > 
> > 
> > I understand, imho it worth to check it once manually, feels to me overkill to add such test, that runs each and everytime, but i will add it. Basically i think that we can take kubevirt as "iron" (assume it works correct, as we are doing usually) but this repo might have bugs, so due to this reason it might be good to have such test, it is just a bit deeper test.
> 
> Please add the test.
> 
> We are extending KubeVirt's behavior, and we need to have an e2e test that shows the integration works as expected.

Agree on adding test.

pkg/testobjects/testobjects.go
```diff
@@ -0,0 +1,130 @@
+package testobjects
```
qinqon, line 1: let's have a `testing` pkg under `pkg/testing` and move this to `pkg/testing/objects.go` if we need we can put sub poackages under `pkg/testing` but one top level testing package per stuff is too much.


Using it like  `testing.DummyVMI` or `ipamtesting.DummyVMI` is ok.

pkg/vminetworkscontroller/vmi_controller.go
```diff
@@ -56,32 +56,58 @@ func (r *VirtualMachineInstanceReconciler) Reconcile(
...
-	if err := r.Client.Get(contextWithTimeout, request.NamespacedName, vm); apierrors.IsNotFound(err) {
-		r.Log.Info("Corresponding VM not found", "vm", request.NamespacedName)
-		ownerInfo = metav1.OwnerReference{APIVersion: vmi.APIVersion, Kind: vmi.Kind, Name: vmi.Name, UID: vmi.UID}
-	} else if err == nil {
-		ownerInfo = metav1.OwnerReference{APIVersion: vm.APIVersion, Kind: vm.Kind, Name: vm.Name, UID: vm.UID}
-	} else {
+	hasVMOwner := false
+	if err := r.Client.Get(contextWithTimeout, request.NamespacedName, vm); err == nil {
+		hasVMOwner = true
+	} else if !apierrors.IsNotFound(err) {
```
qinqon, line 70: I think we can just check the OwnerReferneces from the VMI and look for a VirtualMachien kind, it will more clear.

pkg/vminetworkscontroller/vmi_controller.go
```diff
@@ -56,32 +58,58 @@ func (r *VirtualMachineInstanceReconciler) Reconcile(
...
 		return controllerruntime.Result{}, fmt.Errorf(
 			"error to get VMI %q corresponding VM: %w",
 			request.NamespacedName,
 			err,
 		)
 	}
 
+	if (hasVMOwner && vmi == nil && vm.DeletionTimestamp != nil) ||
+		(!hasVMOwner && (vmi == nil || (vmi.DeletionTimestamp != nil && len(vmi.Status.ActivePods) == 0))) {
```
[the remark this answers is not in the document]
qinqon: Agree, we are not going to understand what we are checking in a weeek or less.

pkg/vminetworkscontroller/vmi_controller.go
```diff
@@ -56,32 +58,58 @@ func (r *VirtualMachineInstanceReconciler) Reconcile(
...
+	} else {
+		ownerInfo = metav1.OwnerReference{
+			APIVersion:         vmi.APIVersion,
+			Kind:               vmi.Kind,
+			Name:               vmi.Name,
+			UID:                vmi.UID,
+			Controller:         ptr.To(true),
+			BlockOwnerDeletion: ptr.To(true),
+		}
+	}
```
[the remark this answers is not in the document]
qinqon: You can generalize this using `client.Object` since all the info is comon

qinqon (changes requested):
first pass

pkg/vminetworkscontroller/vmi_controller.go
oshoval: yes and no, it is doable only if VMI != nil, otherwise we must also try to "get" VM 
so i just kept it as is for simplicity

qinqon: Ahh so we don't have always the owner list since VMI can already being delete, ok, can you add a comment realted to this ?

oshoval: hmm i can but arent we against comments?
anyhow i will add

pkg/testobjects/testobjects.go
maiqueb: since you're going at it - which I wouldn't, I would just have the required helpers directly in the pkg - you'll need to  exclude these files from the binary we ship to customers. 

https://mickey.dev/posts/go-build-tags-testing/

oshoval: I prefer to keep as is now, @maiqueb do you mean the same ?
Reason it that usually unit tests should either duplicate code, or have it on production code and reuse it (as done here)
I think we shoudn't over complicate, and in the other hand i saw many repetitions of the code,
i can also reintroduce the repetitions
wdyt?

oshoval: If there are no objections, i will just drop back to duplication of code as was before this PR
(code was already twice on the required unit test files)

pkg/vminetworkscontroller/vmi_controller.go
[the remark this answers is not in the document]
oshoval: once we add helper here that is unit tested, maybe it makes 
https://github.com/kubevirt/ipam-extensions/pull/40#issuecomment-2242551362
unrequired ?
because the logic if to delete the finalizer from the claim (which what allows the claim to be deleted)
depends 99% solely on this code, this is the case to be precise:
"hasVMOwner && vmi == nil && vm.DeletionTimestamp != nil"

Just something to consider please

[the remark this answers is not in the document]
maiqueb: I think you should add an e2e test w/ the e2e behavior, and I think you should add a unit test here to document what the criteria is for, and why the criteria is what it is.

pkg/testobjects/testobjects.go
maiqueb: I think testing libraries make sense iif they're not shipped w/ the binary. 

IIUC it will unless you exclude it w/ build tags. Maybe I'm wrong.

I'm OK if you drop this whole endeavor and duplicate whatever is required.

FWIW, I really don't know why we have different packages for a vmi controller and for a vm controller. 
We could have a single pkg and 2 different structs, and tests for each of them.

oshoval: > FWIW, I really don't know why we have different packages for a vmi controller and for a vm controller. We could have a single pkg and 2 different structs, and tests for each of them.

btw even if we do that, we still have the webhook that duplicate some, just fyi please

oshoval:
rebased
drafting for now, will update once finished addressing comments

pkg/vminetworkscontroller/vmi_controller_test.go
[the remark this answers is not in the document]
oshoval: since i am dropping the new testobjects pkg as discussed here
https://github.com/kubevirt/ipam-extensions/pull/40#discussion_r1686462944
i can now squash those, so will do it

oshoval:
Addressed comments beside the missing e2e test, now will work on that

oshoval:
Simplified a bit `ownerReferenceFor` to depend solely on the obj argument,
and adapted relevant unit tests to cover better

oshoval:
fixed lint error

oshoval:
some more cosmetics to `ownerReferenceFor`

oshoval:
Added the requested test
Ready for review - thanks

pkg/vminetworkscontroller/vmi_controller.go
```diff
@@ -202,29 +198,42 @@ func (r *VirtualMachineInstanceReconciler) vmiNetworksClaimingIPAM(
...
-	listOpts := []client.ListOption{
-		client.InNamespace(vmiKey.Namespace),
-		ownedByVMLabel(vmiKey.Name),
-	}
-	if err := r.Client.List(context.Background(), ipamClaims, listOpts...); err != nil {
-		return fmt.Errorf("could not get list of IPAMClaims owned by VM %q: %w", vmiKey.String(), err)
+func shouldCleanFinalizers(vmi *virtv1.VirtualMachineInstance, vm *virtv1.VirtualMachine) bool {
+	return (vm != nil && vmi == nil && vm.DeletionTimestamp != nil) ||
+		(vm == nil && (vmi == nil || (vmi.DeletionTimestamp != nil && len(vmi.Status.ActivePods) == 0)))
+}
```
qinqon, line 204: It may be a little more verbose but it will improve readability, is easy to forget about this in a few days, you can even wrap them at helper functions to make it  more clear
```suggestion
func shouldCleanFinalizers(vmi *virtv1.VirtualMachineInstance, vm *virtv1.VirtualMachine) bool {
        // Everything is gone
       if vm == nil && vmi == nil {
           return true
       } 
        
        // VMI is gone and VM is being deleted
        if vmi == nil && vm! == nil && vm.DeletionTimestamp != nil {
           return true
        } 
        
        // Standalone VMI being deleted and no virt-launcher pods
        if vm == nil && vmi != nil && vmi.DeletionTimestamp != nil && len(vmi.Status.ActivePods) == 0 {
            return true
        }
        
        return false
```

pkg/vminetworkscontroller/vmi_controller.go
```diff
@@ -202,29 +198,42 @@ func (r *VirtualMachineInstanceReconciler) vmiNetworksClaimingIPAM(
...
-			}
-		}
+	aPIVersion := obj.GetObjectKind().GroupVersionKind().Group + "/" + obj.GetObjectKind().GroupVersionKind().Version
+	return metav1.OwnerReference{
+		APIVersion:         aPIVersion,
+		Kind:               obj.GetObjectKind().GroupVersionKind().Kind,
+		Name:               obj.GetName(),
+		UID:                obj.GetUID(),
+		Controller:         ptr.To(true),
+		BlockOwnerDeletion: ptr.To(true),
```
qinqon, line 221: There is a tool at controller-runtime to mark owners as controllers, let's use it.
```golang
sigs.k8s.io/pkg/controller/controllerutil.SetControllerReference
```

pkg/vminetworkscontroller/vmi_controller.go
```diff
@@ -202,29 +198,42 @@ func (r *VirtualMachineInstanceReconciler) vmiNetworksClaimingIPAM(
...
+	contextWithTimeout, cancel := context.WithTimeout(ctx, time.Second)
+	defer cancel()
+
+	vm := &virtv1.VirtualMachine{}
+	if err := c.Get(contextWithTimeout, name, vm); err == nil {
+		return vm, nil
+	} else if apierrors.IsNotFound(err) {
+		return nil, nil
+	} else {
+		return nil, fmt.Errorf("error getting VM %q: %w", name, err)
```
qinqon, line 237: Let's start the erro with "failing..." is a veeeery common partern at k8s things.

pkg/vmnetworkscontroller/vm_controller.go
```diff
@@ -0,0 +1,99 @@
...
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	virtv1 "kubevirt.io/api/core/v1"
+)
+
+// VirtualMachineReconciler reconciles a VirtualMachine object
+type VirtualMachineReconciler struct {
```
qinqon, line 23: I am not sure we need to separate Reconcilers, I think we can make one for both VM and VMIs and just check the expected condition `shouldCleanFinalizers` and remove the finalizer is so.

qinqon (changes requested):
I think we can have one controller for both VM and VMI and other nits.

pkg/vmnetworkscontroller/vm_controller.go
oshoval: I prefer please the current separation, easier to maintain and less mental burden imo
each controller does its own,
the VM handles VM events, and VMI handles VMI events, also each one is self contained
and handles it own cases

pkg/claims/claims.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	for _, claim := range ipamClaims.Items {
+		removedFinalizer := controllerutil.RemoveFinalizer(&claim, KubevirtVMFinalizer)
+		if removedFinalizer {
+			if err := c.Update(context.Background(), &claim, &client.UpdateOptions{}); err != nil {
+				return client.IgnoreNotFound(err)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
oshoval: this PR didnt touch this function, i think it we can wait with this change please for a follow-up if its ok

pkg/vminetworkscontroller/vmi_controller.go
oshoval: I prefer please to do it on a follow-up, i did in this PR the minimal changes needed for finalizers
it is better to have PRs focused and not massive

oshoval:
addressed / answered
thx

pkg/vminetworkscontroller/vmi_controller.go
qinqon: Ok, let's do a follow up, it will be something like

```golang
controllerutil.SetControllerReference(vm, vmi, scheme.Scheme)
```

oshoval: yep, it will need to know both cases of VMI owner and VM owner

pkg/claims/claims.go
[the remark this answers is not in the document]
qinqon: This functions was just moved around, let's fix a follow up.

pkg/vminetworkscontroller/vmi_controller.go
```diff
@@ -202,29 +198,59 @@ func (r *VirtualMachineInstanceReconciler) vmiNetworksClaimingIPAM(
...
-	listOpts := []client.ListOption{
-		client.InNamespace(vmiKey.Namespace),
-		ownedByVMLabel(vmiKey.Name),
+func shouldCleanFinalizers(vmi *virtv1.VirtualMachineInstance, vm *virtv1.VirtualMachine) bool {
+	if vm != nil {
+		// VMI is gone and VM is marked for deletion
+		if vmi == nil && vm.DeletionTimestamp != nil {
+			return true
+		}
+		return false
```
qinqon, line 207: We don't need this

qinqon (changes requested):
one nit

pkg/vminetworkscontroller/vmi_controller.go
oshoval: done

test/e2e/persistentips_test.go
```diff
@@ -98,14 +102,60 @@ var _ = Describe("Persistent IPs", func() {
...
+					By("Invoking virtctl stop")
+					output, err := exec.Command("virtctl", "stop", "-n", td.Namespace, vmi.Name).CombinedOutput()
+					Expect(err).NotTo(HaveOccurred(), output)
+
+					By("Ensuring VM is not running")
+					Eventually(testenv.ThisVMI(vmi), 360*time.Second, 1*time.Second).Should(
+						And(
+							Not(testenv.BeCreated()),
+							Not(testenv.BeReady()),
+						))
```
qinqon, line 132: Is more clear this way
```suggestion
						SatisfyAll(
							Not(testenv.BeCreated()),
							Not(testenv.BeReady()),
						))
```

qinqon (changes requested):
just one nit

test/e2e/persistentips_test.go
oshoval: done

qinqon (approved):
/lgtm

oshoval:
fwiw it waits for you, no need hold :)
Thanks

pkg/vminetworkscontroller/vmi_controller_test.go
```diff
@@ -402,3 +536,99 @@ func decorateVMWithUID(uid string, vm *virtv1.VirtualMachine) *virtv1.VirtualMac
...
+	vm.ObjectMeta.Finalizers = []string{metav1.FinalizerDeleteDependents}
+
+	return vm
+}
+
+var _ = Describe("shouldCleanFinalizers", func() {
+	DescribeTable("should determine if finalizers should be cleaned up",
+		func(vmi *virtv1.VirtualMachineInstance, vm *virtv1.VirtualMachine, expectedResult bool) {
+			result := shouldCleanFinalizers(vmi, vm)
+			Expect(result).To(Equal(expectedResult))
```
maiqueb, line 570: nit

```suggestion
			Expect(shouldCleanFinalizers(vmi, vm)).To(Equal(expectedResult))
```

pkg/vminetworkscontroller/vmi_controller_test.go
```diff
@@ -402,3 +536,99 @@ func decorateVMWithUID(uid string, vm *virtv1.VirtualMachine) *virtv1.VirtualMac
...
+func dummyMarkedForDeletionVM(nadName string) *virtv1.VirtualMachine {
+	vm := dummyVM(dummyVMISpec(nadName))
+	vm.DeletionTimestamp = &metav1.Time{Time: time.Now()}
+	vm.ObjectMeta.Finalizers = []string{metav1.FinalizerDeleteDependents}
+
+	return vm
+}
+
+var _ = Describe("shouldCleanFinalizers", func() {
+	DescribeTable("should determine if finalizers should be cleaned up",
```
maiqueb, line 567: would it make sense to re-use the helpers you've introduced above in this test's entries ?

pkg/vminetworkscontroller/vmi_controller_test.go
```diff
@@ -118,20 +123,20 @@ var _ = Describe("vmi IPAM controller", Serial, func() {
 		mgr, err := controllerruntime.NewManager(&rest.Config{}, ctrlOptions)
 		Expect(err).NotTo(HaveOccurred())
 
-		reconcileMachine := NewVMIReconciler(mgr)
+		reconcileVMI := NewVMIReconciler(mgr)
```
maiqueb, line 126: `vmiReconciler` ?

maiqueb (commented):
All I have are opinionated nits. 

Really nice tests you've added.
Let me know if you would prefer to work on those as a follow-up (if at all).

/hold cancel

pkg/vminetworkscontroller/vmi_controller.go
```diff
@@ -202,29 +198,58 @@ func (r *VirtualMachineInstanceReconciler) vmiNetworksClaimingIPAM(
...
+	} else {
+		// VMI is gone
+		if vmi == nil {
+			return true
+		}
+
+		// VMI is marked for deletion, virt-launcher is gone
+		if vmi.DeletionTimestamp != nil && len(vmi.Status.ActivePods) == 0 {
+			return true
+		}
```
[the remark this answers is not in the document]
oshoval: sgtm, @qinqon asked this type of changes basically
@qinqon are you fine with the above suggestion ?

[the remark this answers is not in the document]
qinqon: The original one line condition was not clear, I am ok with whatever make all this conditions clear in a month.

pkg/vminetworkscontroller/vmi_controller_test.go
oshoval: i prefer not, in order to keep the objects as sparse and simple as possible

maiqueb: that argument could be made about the helpers you've added ... Why have them ?...

But ok, we can move on.

oshoval:
addressed / answered comments
thanks

test/e2e/persistentips_test.go
```diff
@@ -130,6 +180,55 @@ var _ = Describe("Persistent IPs", func() {
...
+				Expect(testenv.Client.Delete(context.Background(), vm, foregroundDeleteOptions())).To(Succeed())
+				Consistently(testenv.IPAMClaimsFromNamespace(vm.Namespace)).
+					WithTimeout(time.Minute).
+					WithPolling(time.Second).
+					ShouldNot(BeEmpty())
+
+				By("Remove the finalizer (all the other are already deleted in this stage)")
+				patchData, err := removeFinalizersPatch()
+				Expect(err).NotTo(HaveOccurred())
+				Expect(testenv.Client.Patch(context.TODO(), vmi, client.RawPatch(types.MergePatchType, patchData))).NotTo(HaveOccurred())
```
maiqueb, line 222: You're still not using positive logic. Please use `To(Succeed())` instead of `NotTo(HaveOccurred())` .

maiqueb (commented):
just a single nit from my perspective. 
@qinqon are you OK with this ?

test/e2e/persistentips_test.go
oshoval: missed, done

maiqueb (approved):
/approve

Thank you.

pkg/vminetworkscontroller/vmi_controller.go
```diff
@@ -56,46 +56,42 @@ func (r *VirtualMachineInstanceReconciler) Reconcile(
...
-		ownerInfo = metav1.OwnerReference{APIVersion: vmi.APIVersion, Kind: vmi.Kind, Name: vmi.Name, UID: vmi.UID}
-	} else if err == nil {
-		ownerInfo = metav1.OwnerReference{APIVersion: vm.APIVersion, Kind: vm.Kind, Name: vm.Name, UID: vm.UID}
-	} else {
-		return controllerruntime.Result{}, fmt.Errorf(
-			"error to get VMI %q corresponding VM: %w",
-			request.NamespacedName,
-			err,
-		)
+	vm, err := getOwningVM(ctx, r.Client, request.NamespacedName)
```
oshoval, line 64: thought about something
when getting owning VM, in case we still have the VMI we can validate it is the real owning VM according UID in the owner references,
and not a new VM that was created just after the original VM was deleted (with the same name)

if VMI == nil we can't assert this, but we can look on the ipamclaim ownerRef and compare the UID with the candid owning object (i think this what we should always do, both on VMI controller cases and
maybe also when checking realtionships in VM controller
https://github.com/kubevirt/ipam-extensions/pull/40/files#diff-dc5bb71846a22dd07f268340d8388c08689853df79bf3ae054514d77cecec888R56

PR #48, opened 2024-07-16
WIP: Handle controller restart

oshoval:

Fixes #

Only last commit is relevant
TODO - fix `Update` error handling, it shouldnt return err, but the whole flow
should have another chance to do a successful loop.
We can consider later adding hotunplug logic once we implement hot unplug.
The go routine will look for claims belong to interfaces that are absent and are detached
(according status infoSource multus, but need to make sure about it).

**Checklist**

This checklist is not enforcing, but it's a reminder of items that could be relevant to every PR.
Approvers are expected to review this list.

- [ ] Design: A [design document](https://github.com/kubevirt/community/tree/main/design-proposals) was considered and is present (link) or not required
- [ ] PR: The PR description is expressive enough and will help future contributors
- [ ] Code: [Write code that humans can understand](https://en.wikiquote.org/wiki/Martin_Fowler#code-for-humans) and [Keep it simple](https://en.wikipedia.org/wiki/KISS_principle)
- [ ] Refactor: You have [left the code cleaner than you found it (Boy Scout Rule)](https://learning.oreilly.com/library/view/97-things-every/9780596809515/ch08.html)
- [ ] Upgrade: Impact of this change on upgrade flows was considered and addressed if required
- [ ] Testing: New code requires [new unit tests](https://github.com/kubevirt/kubevirt/blob/main/docs/reviewer-guide.md#when-is-a-pr-good-enough). New features and bug fixes require at least on e2e test
- [ ] Documentation: A [user-guide update](https://github.com/kubevirt/user-guide/) was considered and is present (link) or not required. You want a user-guide update if it's a user facing feature / API change.
- [ ] Community: Announcement to [kubevirt-dev](https://groups.google.com/g/kubevirt-dev/) was considered

**Release note**:

```release-note

```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -17,9 +17,13 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"crypto/tls"
 	"flag"
+	"fmt"
 	"os"
+	"strings"
+	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
@@ -29,6 +33,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 
@@ -41,9 +46,16 @@ import (
 
 	virtv1 "kubevirt.io/api/core/v1"
 
+	"k8s.io/apimachinery/pkg/types"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
 	ipamclaimsapi "github.com/k8snetworkplumbingwg/ipamclaims/pkg/crd/ipamclaims/v1alpha1"
 	nadv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 
+	"github.com/kubevirt/ipam-extensions/pkg/claims"
 	"github.com/kubevirt/ipam-extensions/pkg/ipamclaimswebhook"
 	"github.com/kubevirt/ipam-extensions/pkg/vminetworkscontroller"
 	"github.com/kubevirt/ipam-extensions/pkg/vmnetworkscontroller"
@@ -176,6 +188,13 @@ func main() {
 		&webhook.Admission{Handler: ipamclaimswebhook.NewIPAMClaimsValet(mgr)},
 	)
 
+	go func() {
+		if err := purgeFinalizerForMissingOwners(mgr.GetClient()); err != nil {
+			setupLog.Error(err, "problem purging finalizers")
+			os.Exit(1)
+		}
+	}()
+
 	setupLog.Info("starting manager")
 	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
 		setupLog.Error(err, "problem running manager")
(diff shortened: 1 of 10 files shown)


maiqueb:
Are you sure this is needed ? 

I'm almost sure controller runtime's controller are triggered when the manager starts. And on startup, it should process every item whose key it doesn't have cached (which are all keys).

oshoval:
> Are you sure this is needed ?
> 
> I'm almost sure controller runtime's controller are triggered when the manager starts. And on startup, it should process every item whose key it doesn't have cached (which are all keys).

Controller runtime does fire events for each object that exists upon manager start, but not for objects that does not exists,
which is why we have an issue.

I managed to simulate the following case where it happened,
Deployed KIC, 
Created a VM
Undeploy KIC - simulating for example a case where the node where the manager runs was rebooted, hence the pod is NA
Delete the VM (background) 
Redeploy KIC
Watch that the ipam claim zombie is left forever
and with this PR it is deleted.

maiqueb:
> > Are you sure this is needed ?
> > I'm almost sure controller runtime's controller are triggered when the manager starts. And on startup, it should process every item whose key it doesn't have cached (which are all keys).
> 
> Controller runtime does fire events for each object that exists upon manager start, but not for objects that does not exists, which is why we have an issue.
> 
> I managed to simulate the following case where it happened:
> 
> * Deployed KIC

What is KIC ? Please use less 3 letter acronyms / stop adding them. It's not clear what you mean.

> * Created a VM
> * Undeploy KIC - simulating for example a case where the node where the manager runs was rebooted, hence the pod is NA
> * Delete the VM (background delete)
> * Redeploy KIC
> * Saw that the IPAM Claim zombie is left forever

I'd say the admin is welcome to delete them manually.

If they don't want to have there, they should not have deleted the claims manager in the first place.

> 
> and with this PR it is deleted.

Let me think on this ... I am not sure we actually want to able to address this situation.

oshoval:
> > > Are you sure this is needed ?
> > > I'm almost sure controller runtime's controller are triggered when the manager starts. And on startup, it should process every item whose key it doesn't have cached (which are all keys).
> > 
> > 
> > Controller runtime does fire events for each object that exists upon manager start, but not for objects that does not exists, which is why we have an issue.
> > I managed to simulate the following case where it happened:
> > 
> > * Deployed KIC
> 
> What is KIC ? Please use less 3 letter acronyms / stop adding them. It's not clear what you mean.
Kubevirt ipam controller

> 
> > * Created a VM
> > * Undeploy KIC - simulating for example a case where the node where the manager runs was rebooted, hence the pod is NA
> > * Delete the VM (background delete)
> > * Redeploy KIC
> > * Saw that the IPAM Claim zombie is left forever
> 
> I'd say the admin is welcome to delete them manually.

But as long as he doesnt they might hold IPs.

> 
> If they don't want to have there, they should not have deleted the claims manager in the first place.

Maybe the node crashed, or the pod was evicted, and during that, it happened, it wasn't deliberately done by the admin.

> 
> > and with this PR it is deleted.
> 
> Let me think on this ... I am not sure we actually want to able to address this situation.

Sure thanks

oshoval:
Btw one thing to consider please,
this new function, with some adaption can be changed to help us handle unplug,
or just present it once we handle hot unplug and then reconsider covering this case

in brief, for hot unplug as we spoke once, we can add a code that reads the claims every now and then, and check what can be deleted due to hot unplugs, and then it can also delete those zombies.

maiqueb:
> Btw one thing to consider please, this new function, with some adaption can be used to help us handle unplug, or just present it once we handle hot unplug and then reconsider covering this case
> 
> in brief, for hot unplug as we spoke once, we can add a code that reads the claims every now and then, and check what can be deleted due to hot unplugs, and then it can also delete those zombies.

Let's take it into account when designing plug / unplug support.

oshoval:
sure closing

oshoval:
If k8s supported conditional finalizers,
i.e a finalizer that is "enabled" only if the owner exists,
we wouldn't even need to handle this issue, as it was done automatically
i wonder if there are more interesting use cases / if it worth to ask the community if it worth suggesting such thing?
(not atm of course, as stuff like this consume time)

PR #68, opened 2024-09-16
cluster: Use --opt-out-kv-ipam

oshoval:
Make cluster-sync going to install it anyhow.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/hack/cluster.sh
+++ b/hack/cluster.sh
@@ -52,7 +52,7 @@ function up() {
     kind delete cluster --name $cluster_name
     (
         cd ${OVN_KUBERNETES_DIR}
-        ./contrib/kind.sh --local-kind-registry ${KIND_ARGS} -cn ${cluster_name}
+        ./contrib/kind.sh --local-kind-registry ${KIND_ARGS} -cn ${cluster_name} --opt-out-kv-ipam
     )
 }
 


oshoval:
```
#8 13.85 >>> Librepo error: Cannot download repomd.xml: Cannot download repodata/repomd.xFailed to download metadata (metalink: "https://mirrors.fedoraproject.org/metalink?repo=rawhide&arch=x86_64") for repository "rawhide"
#8 13.85  Librepo error: Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
```
some hiccup while dnf install

oshoval:
https://pagure.io/fedora-infrastructure/issue/12183
Fedora issue here

oshoval:
interesting, flake on unit test
i think it is not related to this PR, as it changes only e2e

maiqueb:
> interesting, flake on unit test (unless something was changed and it will fail consistently) https://github.com/kubevirt/ipam-extensions/actions/runs/10900433841/job/30248013978?pr=68 i think it is not related to this PR, as it changes only e2e

Hm, it's the order right ? Do we really need the order of this to be this static ? 
If not, we can just use consist of instead of equal in the tests.
@qinqon

qinqon:
> > interesting, flake on unit test (unless something was changed and it will fail consistently) https://github.com/kubevirt/ipam-extensions/actions/runs/10900433841/job/30248013978?pr=68 i think it is not related to this PR, as it changes only e2e
> 
> Hm, it's the order right ? Do we really need the order of this to be this static ? If not, we can just use consist of instead of equal in the tests. @qinqon

We are not crafting the jsonpatch ourselves, we just pass new and old pods to `admission.PatchResponseFromRaw` so this is supose to be fine, problem is that at slices order matter when we compare things at tests, even if we compare them as a json.

We have to fix the test so the test case instruct to ignore the order there.

PR #70, opened 2024-09-22
pod mutator, unit tests: Fix flaky test

oshoval:
Since the patches might appear in different order on some cases
we should use `ConsistOf` on those cases, and keep `Equal` by default,
since the patches are applied in order.
Example
https://github.com/kubevirt/ipam-extensions/actions/runs/10900433841/job/30248013978?pr=68

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/pkg/ipamclaimswebhook/podmutator_test.go
+++ b/pkg/ipamclaimswebhook/podmutator_test.go
@@ -9,6 +9,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/types"
 
 	"gomodules.xyz/jsonpatch/v2"
 
@@ -38,7 +39,8 @@ type testConfig struct {
 	inputVMI                  *virtv1.VirtualMachineInstance
 	inputNADs                 []*nadv1.NetworkAttachmentDefinition
 	inputPod                  *corev1.Pod
-	expectedAdmissionResponse admission.Response
+	expectedAdmissionResponse admissionv1.AdmissionResponse
+	expectedAdmissionPatches  types.GomegaMatcher
 }
 
 func TestController(t *testing.T) {
@@ -105,11 +107,12 @@ var _ = Describe("KubeVirt IPAM launcher pod mutato machine", Serial, func() {
 
 		ipamClaimsManager := NewIPAMClaimsValet(mgr)
 
-		Expect(
-			ipamClaimsManager.Handle(context.Background(), podAdmissionRequest(config.inputPod)),
-		).To(
-			Equal(config.expectedAdmissionResponse),
-		)
+		result := ipamClaimsManager.Handle(context.Background(), podAdmissionRequest(config.inputPod))
+
+		Expect(result.AdmissionResponse).To(Equal(config.expectedAdmissionResponse))
+		if config.expectedAdmissionPatches != nil {
+			Expect(result.Patches).To(config.expectedAdmissionPatches)
+		}
 	},
 		Entry("pod not beloging to a VM and not requesting secondary "+
 			"attachments and no primary user defined network is accepted", testConfig{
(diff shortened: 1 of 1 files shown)


pkg/ipamclaimswebhook/podmutator_test.go
```diff
@@ -105,11 +105,10 @@ var _ = Describe("KubeVirt IPAM launcher pod mutato machine", Serial, func() {
...
 
-		Expect(
-			ipamClaimsManager.Handle(context.Background(), podAdmissionRequest(config.inputPod)),
-		).To(
-			Equal(config.expectedAdmissionResponse),
-		)
+		result := ipamClaimsManager.Handle(context.Background(), podAdmissionRequest(config.inputPod))
+
+		Expect(result.Patches).To(ConsistOf(config.expectedAdmissionResponse.Patches))
+		Expect(result.AdmissionResponse).To(Equal(config.expectedAdmissionResponse.AdmissionResponse))
```
maiqueb, line 111: maybe switch the order of these two ? 
@qinqon you said something about the order not mattering only for a *subset* of the tests. Could you elaborate ? 

Can this PR be merged ?

oshoval: > maybe switch the order of these two ?

the only benefit is fail fast if the AdmissionResponse is different
i think it is a nit but switched (done)

imho the PR can be merged, but maybe you wait for Quique

pkg/ipamclaimswebhook/podmutator_test.go
```diff
@@ -105,11 +105,10 @@ var _ = Describe("KubeVirt IPAM launcher pod mutato machine", Serial, func() {
...
 
-		Expect(
-			ipamClaimsManager.Handle(context.Background(), podAdmissionRequest(config.inputPod)),
-		).To(
-			Equal(config.expectedAdmissionResponse),
-		)
+		result := ipamClaimsManager.Handle(context.Background(), podAdmissionRequest(config.inputPod))
+
+		Expect(result.AdmissionResponse).To(Equal(config.expectedAdmissionResponse.AdmissionResponse))
+		Expect(result.Patches).To(ConsistOf(config.expectedAdmissionResponse.Patches))
```
qinqon, line 111: We should do this per test not for all the test, since depending on the test it's important or not.

One way is to refactor the test so the `expectedPatches` is `gomega.Matcher` so we can do 

```golang
expectedPatch: ConsistOf(....)
```

qinqon (changes requested):

pkg/ipamclaimswebhook/podmutator_test.go
oshoval: why? the order doesnt matter never on the patches field, is can always be ConsistOf,
consider please to keep it simple as it is now

oshoval: I understand, i think it is wrong, because we don't have it right now, and if we have then we can fix
atm we don't see we going to have such case, anyhow will change

pkg/ipamclaimswebhook/podmutator_test.go
```diff
@@ -39,6 +40,7 @@ type testConfig struct {
 	inputNADs                 []*nadv1.NetworkAttachmentDefinition
 	inputPod                  *corev1.Pod
 	expectedAdmissionResponse admission.Response
+	matcherFunc               interface{}
```
qinqon, line 43: Let's maek this more explicit and not use a very opaque `interface{}` function, we can split the expectations for the response into two:

```golang

admissionv1 "k8s.io/api/admission/v1"
type type testConfig struct { 
....
   expectedAdmissionResponse dmissionv1.AdmissionResponse
   expectedAdmissionPatches  types.GomegaMatcher
}
```

Then this can be use like 

```golang
		testConfig{
			inputVM:  dummyVM(nadName),
			inputVMI: dummyVMI(nadName),
			inputNADs: []*nadv1.NetworkAttachmentDefinition{
				dummyNAD(nadName),
			},
			inputPod: &corev1.Pod{},
			expectedAdmissionResponse: admissionv1.AdmissionResponse{
					Allowed:   true,
					PatchType: &patchType,
				},
			expectedAdmissionPatches: Equals([]jsonpatch.JsonPatchOperation{
				{
					Operation: "add",
					Path:      "/metadata/annotations/k8s.ovn.org~1primary-udn-ipamclaim",
					Value:     "vm1.podnet",
				},
				{
					Operation: "replace",
					Path:      "/metadata/annotations/k8s.v1.cni.cncf.io~1networks",
					Value:     "[{\"name\":\"supadupanet\",\"namespace\":\"ns1\",\"ipam-claim-reference\":\"vm1.randomnet\"}]",
				},
			}),
		},
```

qinqon (changes requested):
I was refereing on spliting the expectation on the response

pkg/ipamclaimswebhook/podmutator_test.go
oshoval: done, ptal

PR #71, opened 2024-09-26
e2e, persistent-ip: Add primary-UDN tests

RamLavi:
This PR adds the needed tests needed in order to cover ipam-extentions operation on VMI/VMs using primary-UDN.

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/test/e2e/persistentips-primary_test.go
+++ b/test/e2e/persistentips-primary_test.go
@@ -1,0 +1,396 @@
+/*
+ * This file is part of the KubeVirt project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2024 Red Hat, Inc.
+ *
+ */
+
+package e2e
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os/exec"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	nadv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+
+	kubevirtv1 "kubevirt.io/api/core/v1"
+
+	testenv "github.com/kubevirt/ipam-extensions/test/env"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("Persistent IPs on Primary UDN interface", func() {
+	var failureCount int = 0
+	JustAfterEach(func() {
+		if CurrentSpecReport().Failed() {
+			failureCount++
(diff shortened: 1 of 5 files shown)


RamLavi:
Added [commit](https://github.com/kubevirt/ipam-extensions/pull/71/commits/a1df094dbdcca71b5b2859c595bb7dbb07163424) to prevent the e2e from failing on restart scenario. 
@qinqon perhaps this is also needed in the secondary case?

test/e2e/persistentips-primary_test.go
```diff
@@ -116,6 +116,8 @@ var _ = Describe("Persistent IPs on Primary UDN interface", func() {
 					WithTimeout(5 * time.Minute).
 					Should(testenv.ContainConditionVMIReady())
 
+				Expect(testenv.Client.Get(context.Background(), client.ObjectKeyFromObject(vmi), vmi)).To(Succeed())
```
oshoval, line 119: please consider making it part of the Eventually block
(also on the other places)
this what we are usually on kubevirt fwiw

RamLavi: I don't think we need eventually, because this call is right after we wait for the VMI to be ready

oshoval: you already fetch the latest vmi as part of `ThisVMI` (implicit Eventually happens for what you need)
so you can just use that value externally

func ThisVMI(vmi *kubevirtv1.VirtualMachineInstance) func() (*kubevirtv1.VirtualMachineInstance, error) {
	return func() (p *kubevirtv1.VirtualMachineInstance, err error) {
		p = &kubevirtv1.VirtualMachineInstance{}
		err = Client.Get(context.Background(), client.ObjectKeyFromObject(vmi), p)
		if apierrors.IsNotFound(err) {
			return nil, nil
		}
		return
	}
}

oshoval:
> Added [commit](https://github.com/kubevirt/ipam-extensions/pull/71/commits/a1df094dbdcca71b5b2859c595bb7dbb07163424) to prevent the e2e from failing on restart scenario. @qinqon perhaps this is also needed in the secondary case?

for now they are stable, i wouldnt suggest to add unless we see either by code reading or by a failure that it is really needed
not sure if the exact code / flow is there (well you know better as you derived from there)

test/e2e/persistentips-primary_test.go
RamLavi: true, but this change has bigger scope than what I want to achieve in this PR. 
But we should definitely should do it on a follow up PR (change ThisVMI + remove all the "fetch" that is used before the assertion)

oshoval: honestly i dont understand why to not add 
`vmi = testenv.ThisVMI(vmi)` in the first line of the Eventually and thats all

RamLavi:
> > Added [commit](https://github.com/kubevirt/ipam-extensions/pull/71/commits/a1df094dbdcca71b5b2859c595bb7dbb07163424) to prevent the e2e from failing on restart scenario. @qinqon perhaps this is also needed in the secondary case?
> 
> for now they are stable, i wouldnt suggest to add unless we see either by code reading or by a failure that it is really needed not sure if the exact code / flow is there (well you know better as you derived from there)

ah no no ... my bad. I need to remove that commit altogether..

RamLavi:
[Change](https://github.com/kubevirt/ipam-extensions/compare/a1df094dbdcca71b5b2859c595bb7dbb07163424..d3f56d5166d7262576c1bc997db30d8db24b8136): Rebase

test/e2e/persistentips-primary_test.go
RamLavi: I intend to do so, but in a larger scope, in a follow up PR

qinqon:
@RamLavi let's refactor current tests a little to parameterize role and VM's interface (so we can pass the proper primary UDN network binding).

RamLavi:
closing in favor of https://github.com/kubevirt/ipam-extensions/pull/72

PR #77, opened 2024-12-11
template: specify priority class

maiqueb:
This PR specifies the deployment's priority class, thus assuring the component is not preempted for other workloads with lesses priority. 

Fixes #

**Release note**:

```release-note
Specify the kubevirt-cluster-critical priorityClass for this deployment. It requires KubeVirt to be deployed.
```

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -91,5 +91,6 @@ spec:
           requests:
             cpu: 10m
             memory: 64Mi
+      priorityClassName: system-cluster-critical
       serviceAccountName: manager
       terminationGracePeriodSeconds: 10
--- a/dist/install.yaml
+++ b/dist/install.yaml
@@ -259,6 +259,7 @@ spec:
         - mountPath: /tmp/k8s-webhook-server/serving-certs
           name: cert
           readOnly: true
+      priorityClassName: system-cluster-critical
       securityContext:
         runAsNonRoot: true
         seccompProfile:


oshoval (commented):
what happen if someone tries to deploy it without kubevirt ?
(and the PC doesnt exists)
well he cant anyhow because it listens to VM / VMI i guess so it is non issue

maiqueb:
> what happen if someone tries to deploy it without kubevirt ? (and the PC doesnt exists) well he cant anyhow because it listens to VM / VMI i guess so it is non issue

Yeah, that's kind of the point ... This is a project in the kubevirt org ... 

I guess we should add to the docs a note indicating this must be used w/ kubevirt.

Let me check what happens though ...

oshoval:
Please dont over invest on that, it also happens on KubeSecondaryDNS, and we concluded there it is fine
release note is more than enough imho

oshoval:
btw are you sure you want to use this pc ?
i remember that we used the system critical pc and the other one that k8s gives for the components (forgot the name),
we considered components that are mandatory for kubevirt to have higher prio than kubevirt
if it was a DS it was highest prio pc, 
if it was a Deployment that could be on any node, it was the one below that, but higher than kubevirt
Petr was in the loop, in case i am wrong

maiqueb:
> btw are you sure you want to use this pc ? i remember that we used `system-node-critical` and `system-node-critical` built in pcs for sig network components.
> 
> We considered components that are mandatory for kubevirt to have higher prio than kubevirt, if it was a DaemonSet it was highest prio pc, system-critical if it was a Deployment that could be on any node, it was node-critical Petr was in the loop, in case i am wrong

So you're suggesting we use `system-node-critical` instead ?

oshoval:
yes, wdyt?
if you want to make sure Petr is best with that

phoracek:
For a deployment that other components relay on, use cluster critical. We don't have to be node critical - it is ok if the pod moves to a different node.

phoracek:
@oshoval 

> We considered components that are mandatory for kubevirt to have higher prio than kubevirt, if it was a DaemonSet it was highest prio pc, system-critical if it was a Deployment that could be on any node, it was node-critical Petr was in the loop, in case i am wrong

I don't remember / understand this. Why would we set node critical for a deployment?

oshoval:
I have a typo there / mix

I remember there are 2 classes, highest which is only for stuff that should be all on nodes
and one below that just need to be one stuff like Deployment that are allowed to switch but critical to have, such as KMP
forgot who is who

maiqueb:
@oshoval please take another look

oshoval:
Nit:
Release note: "It requires KubeVirt to be deployed." still true but not due to the PC / this PR
Thanks

PR #88, opened 2025-01-23
config/manager, ipam-ext Deployment: Remove limits, Increase reqeusts

RamLavi:
control-plane components such as ipam-ext Deployment should not have memory/cpu limit.
This PR removes the limits.
Also, Increasing memory request to those [requested](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/main/config/release/kubemacpool.yaml#L301) by similar components like kubemacpool.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -87,7 +87,7 @@ spec:
         resources:
           requests:
             cpu: 10m
-            memory: 64Mi
+            memory: 128Mi
       priorityClassName: system-cluster-critical
       serviceAccountName: manager
       terminationGracePeriodSeconds: 10
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -85,9 +85,6 @@ spec:
           initialDelaySeconds: 5
           periodSeconds: 10
         resources:
-          limits:
-            cpu: 500m
-            memory: 128Mi
           requests:
             cpu: 10m
             memory: 64Mi
--- a/dist/install.yaml
+++ b/dist/install.yaml
@@ -244,9 +244,6 @@ spec:
           initialDelaySeconds: 5
           periodSeconds: 10
         resources:
-          limits:
-            cpu: 500m
-            memory: 128Mi
           requests:
             cpu: 10m
             memory: 64Mi
--- a/dist/install.yaml
+++ b/dist/install.yaml
@@ -246,7 +246,7 @@ spec:
         resources:
           requests:
             cpu: 10m
-            memory: 64Mi
+            memory: 128Mi
         securityContext:
           allowPrivilegeEscalation: false
           capabilities:


config/manager/manager.yaml
```diff
@@ -85,9 +85,6 @@ spec:
           initialDelaySeconds: 5
           periodSeconds: 10
         resources:
-          limits:
-            cpu: 500m
-            memory: 128Mi
           requests:
             cpu: 10m
             memory: 64Mi
```
maiqueb, line 90: Could we bump this to 128 ?... 

Would that make sense ? How much do we use in Kubemacpool ?

RamLavi: [20Mi](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/main/config/release/kubemacpool.yaml#L325C13-L325C25)

RamLavi: opps, my bad
[100Mi](https://github.com/k8snetworkplumbingwg/kubemacpool/blob/main/config/release/kubemacpool.yaml#L325C13-L325C25)

maiqueb: ok, sold ... Let's stick to this number, and implement paging in a follow-up. I'll create a ticket.

RamLavi: I think 128Mi is not too low, not too high.. why not.

maiqueb:
Oh you need to run `make install` and commit the changes you'll see in the deployment manifest. 
Please make sure to define the variable with the IMG name.

config/manager/manager.yaml
RamLavi: Added..

RamLavi:
> Oh you need to run `make install` and commit the changes you'll see in the deployment manifest. Please make sure to define the variable with the IMG name.

there is no `make install`:
```
ipam-extensions (remove_limit) $ make install
make: *** No rule to make target 'install'.  Stop.
```

I did do `make build manifest` - but it didn't seem to to anything, so I figured you don't save the manifest created by kustomize. is that the case? rechecking...

maiqueb:
> > Oh you need to run `make install` and commit the changes you'll see in the deployment manifest. Please make sure to define the variable with the IMG name.
> 
> there is no `make install`:
> 
> ```
> ipam-extensions (remove_limit) $ make install
> make: *** No rule to make target 'install'.  Stop.
> ```
> 
> I did do `make build manifest` - but it didn't seem to to anything, so I figured you don't save the manifest created by kustomize. is that the case? rechecking...

Run `IMG=ghcr.io/kubevirt/ipam-controller:latest make build-installer`

RamLavi:
CI tests fails.. weird. checking locally

maiqueb (approved):
/approve

PR #90, opened 2025-02-04
nad: Skip NADs with empty config

oshoval:
Until now, a NAD with empty config, was considered invalid,
causing an error / rejecting it on the webhook.
Istio creates such NADs, so we can't consider them as fatal,
once we list all the NADs in a namespace.
Otherwise the webhook will block VM creation [1] in case it uses
such NAD.

Therefore just ignore those.

```
[1] virtualmachine-controller    Error creating pod: admission webhook "ipam-claims.k8s.cni.cncf.io" denied the request: failed to extract CNI configuration from NAD: unexpected end of JSON input
```

Example of resource that create such NAD:
```
apiVersion: maistra.io/v1
kind: ServiceMeshMemberRoll
metadata:
  name: default
  namespace: istio-system
spec:
  members:
  - non-udn-ns
```

Fixes #

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/config/types.go
+++ b/pkg/config/types.go
@@ -21,6 +21,9 @@ type RelevantConfig struct {
 
 func NewConfig(nadSpec string) (*RelevantConfig, error) {
 	nadConfig := &RelevantConfig{}
+	if nadSpec == "" {
+		return nadConfig, nil
+	}
 	if err := json.Unmarshal([]byte(nadSpec), nadConfig); err != nil {
 		return nil, fmt.Errorf("failed to extract CNI configuration from NAD: %w", err)
 	}


oshoval:
Hi @maiqueb 
I can just change `NewConfig` but it will ignore empty NADs even on NADs that were supposed to belong to OVN-K
(unless for example we filter by istio label, but since this label might get changed and due to the relative complexity for just this purpose, i don't think this is good idea)
Would you prefer the current solution or just to change `NewConfig`
Once we converge i can add UT, thanks

maiqueb:
> Hi @maiqueb I can just change `NewConfig` but it will ignore empty NADs even on NADs that were supposed to belong to OVN-K (unless for example we filter by istio label, but since this label might get changed and due to the relative complexity for just this purpose, i don't think this is good idea) Would you prefer the current solution or just to change `NewConfig` ? Once we converge i can add UT, thanks

@oshoval how can an empty NAD belong to OVN-K ? 

For a NAD to be associated w/ a plugin, it must feature the name of the plugin in their `type`. If it's empty, it can't belong to it. Hence, nothing to do.

oshoval:
> @oshoval how can an empty NAD belong to OVN-K ?

of course it shouldn't, i mean just because of a bug (which was the original reason we added it in the first place)
anyhow will consolidate

maiqueb (approved):
/lgtm

maiqueb:
> /lgtm

/approve

oshoval:
I will just issue a release and we make sure it works with it, as it shouldn't be risky

oshoval:
btw note please that sometimes doing double tag on the same hash cause mistakes on CNAO
(some of them were solved, some not sure yet)
*   05b47ce (tag: v0.2.0, tag: v0.1.11-alpha) Merge pull request #88 from RamLavi/remove_limit

PR #91, opened 2025-02-04
Kustomize openshift

maiqueb:
This PR provides an installation manifest to deploy the component correctly in OpenShift.

Fixes #

**Release note**:

```release-note
Provide an OpenShift deployment manifest.
```


maiqueb:
cc @oshoval

config/openshift/kustomization.yaml
```diff
@@ -0,0 +1,43 @@
+# Adds namespace to all resources.
+namespace: openshift-kubevirt-ipam-controller-system
```
maiqueb, line 2: @oshoval / @qinqon should this be `openshift-cnv` ??

I don't think so, since this method of installation is just for testing this works without having to grab the whole CNV bundle (i.e. smoke tests, assert this deploys, etc).

config/openshift/kustomization.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+# Value of this field is prepended to the
+# names of all resources, e.g. a deployment named
+# "wordpress" becomes "alices-wordpress".
+# Note that it should also match with the prefix (text before '-') of the namespace
+# field above.
+namePrefix: kubevirt-ipam-controller-
```
maiqueb, line 9: I think this must match w/ the namespace name we have above (line 2).

config/openshift/kustomization.yaml
oshoval: i think you can just do whatever ns yo uwant
if it helps, this worked for me i think
https://gist.github.com/oshoval/d5db815ebea60eb1777a838bee0530b9#file-gistfile1-txt-L14

might just need to add the scc fix for 4.19

config/openshift/kustomization.yaml
oshoval: i would say dont do kustomize at all, provide static manifest imho

oshoval:
if you insist using kustomize
maybe you can give AI the input manifest and output and it will generate you the kustomize

PR #101, opened 2025-05-18
Add Network policies

ormergi:
In a scenario where a cluster networking has been hardened and futures
a default deny-all network policy that affects kubevirt-ipam-controller namespace,
it will not be able to operate properly.

This PR introduce network-policies for the project allowing kubevirt-ipam-controller to operate,
complying with restrictions in form of deny all network-policy.

In order to have coverage for the introduced network-policies, and additional deny-all network-policies is installed as part of the cluster-sync flow.
Having a deny-all network-policy installed, allow the introduced network-policies to be tested in a development and test environments, locally and CI.

**Release note**:

```release-note
config: Introduce network-policies enable kubevirt-ipam-controller operate in restricted env
```

### DIFF
--- a/config/networkpolicy/allow-egress-dns.yaml
+++ b/config/networkpolicy/allow-egress-dns.yaml
@@ -1,0 +1,24 @@
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-egress-to-dns
+spec:
+  podSelector:
+    matchLabels:
+      app: ipam-virt-workloads
+  policyTypes:
+  - Egress
+  egress:
+    - to:
+      - namespaceSelector:
+          matchLabels:
+            kubernetes.io/metadata.name: kube-system
+        podSelector:
+          matchLabels:
+            k8s-app: kube-dns
+      ports:
+        - protocol: TCP
+          port: dns-tcp
+        - protocol: UDP
+          port: dns
--- a/config/networkpolicy/allow-egress-kube-apiserver.yaml
+++ b/config/networkpolicy/allow-egress-kube-apiserver.yaml
@@ -1,0 +1,15 @@
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-egress-to-api-server
+spec:
+  podSelector:
+    matchLabels:
+      app: ipam-virt-workloads
+  policyTypes:
+  - Egress
+  egress:
+  - ports:
+    - protocol: TCP
+      port: 6443
--- a/config/networkpolicy/allow-ingress-to-service.yaml
+++ b/config/networkpolicy/allow-ingress-to-service.yaml
@@ -1,0 +1,15 @@
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-ingress-to-ipam-ext-webhook
+spec:
+  podSelector:
+    matchLabels:
+      app: ipam-virt-workloads
+  policyTypes:
+    - Ingress
+  ingress:
+  - ports:
+    - protocol: TCP
+      port: 9443
--- a/hack/install-deny-all-net-pol.sh
+++ b/hack/install-deny-all-net-pol.sh
@@ -1,0 +1,37 @@
+#!/bin/bash -ex
+#
+# Copyright 2025 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# This script install deny-all NetworkPolicy that affects kubevirt-ipam-controller namespace.
+
+readonly ns="$(${KUBECTL} get pod -l app=ipam-virt-workloads -A -o=custom-columns=NS:.metadata.namespace --no-headers | head -1)"
+[[ -z "${ns}" ]] && echo "FATAL: kubevirt-ipam-controller pods not found. Make sure kubevirt-ipam-controller is installed" && exit 1
+
+readonly np_name="deny-all"
+${KUBECTL} -n "${ns}" get networkpolicy "${np_name}" &> /dev/null ||
+  cat <<EOF | ${KUBECTL} -n "${ns}" apply -f -
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: ${np_name}
+spec:
+  podSelector: {}
+  policyTypes:
+  - Ingress
+  - Egress
+  ingress: []
+  egress: []
+EOF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -18,6 +18,7 @@ resources:
 - ../manager
 - ../webhook
 - ../certmanager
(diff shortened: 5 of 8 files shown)


config/networkpolicy/allow-egress-dns.yaml
```diff
@@ -0,0 +1,24 @@
...
+  podSelector:
+    matchLabels:
+      app: ipam-virt-workloads
+  policyTypes:
+  - Egress
+  egress:
+    - to:
+      - namespaceSelector:
+          matchLabels:
+            kubernetes.io/metadata.name: kube-system
```
maiqueb, line 16: this will need to be kustomized so we can reference a different namespace downstream, right ?

config/networkpolicy/allow-egress-kube-apiserver.yaml
```diff
@@ -0,0 +1,15 @@
...
+spec:
+  podSelector:
+    matchLabels:
+      app: ipam-virt-workloads
+  policyTypes:
+  - Egress
+  egress:
+  - ports:
+    - protocol: TCP
+      port: 6443
```
maiqueb, line 15: can't we be more specific than this ?

config/networkpolicy/allow-ingress-to-service.yaml
```diff
@@ -0,0 +1,15 @@
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-ingress-to-ipm-ext-service
```
maiqueb, line 5:
```suggestion
  name: allow-ingress-to-ipam-ext-service
```

maiqueb: maybe even be clearer and refer this is to allow contacting the webhook

hack/install-deny-all-net-pol.sh
```diff
@@ -0,0 +1,37 @@
...
+#
+
+# This script install deny-all NetworkPolicy that affects kubevirt-ipam-controller namespace.
+
+readonly ns="$(${KUBECTL} get pod -l app=ipam-virt-workloads -A -o=custom-columns=NS:.metadata.namespace --no-headers | head -1)"
+[[ -z "${ns}" ]] && echo "FATAL: kubevirt-ipam-controller pods not found. Make sure kubevirt-ipam-controller is installed" && exit 1
+
+readonly np_name="deny-all"
+${KUBECTL} -n "${ns}" get networkpolicy "${np_name}" ||
+  cat <<EOF | ${KUBECTL} -n "${ns}" create -f -
```
maiqueb, line 25: let's make this idempotent by using apply instead. 

I think this'll fail if sync is invoked twice.

maiqueb (changes requested):
Thank you.

config/networkpolicy/allow-egress-kube-apiserver.yaml
ormergi: Unfortunately we cant, network policy does not allow you to select endpoints on the host network such as kube-apiserver ones.

hack/install-deny-all-net-pol.sh
ormergi: You totally right, it misses some bits that makes it idempotent.

ormergi:
[Changes](https://github.com/kubevirt/ipam-extensions/compare/ffff69fe6abb035f614c4716ef06c8b367f0ebcf..195ebef0c886f7acdbf447086f2efc675183faf2): resolve https://github.com/kubevirt/ipam-extensions/pull/101#pullrequestreview-2849847298 

@maiqueb PTAL

hack/install-deny-all-net-pol.sh
ormergi: I verified it works when sync invoked multiple times, missing part was passing stdout/err to /dev/null.

ormergi: Its possible, but I dont think we need to use apply because this policy is not expected to changed.
Note that it wont create the object if its already exist.

maiqueb: ok, then it's just my OCD at play ... still I got to ask, why use create vs apply ?

ormergi: I dont have strong opinions in this particular case. I changed it to use apply, PTAL.

config/networkpolicy/allow-egress-dns.yaml
maiqueb: Well, IDK. 

What I'd do, is try to fix this in kustomize, because CNAO does some really weird things with some scripts to template the components. I'd reaaaaaaally avoid trying to solve the problem in that scope. Hence, I'd try to solve the problem with better tools whenever they exist. Luckily ipam-extensions is kustomize aware.

oshoval: i would suggest gladly to keep CNAO do what needed, as we already did for deploying ipam on OpenShift
we are doing the same for all components, it is easy and straight forward
since the values are binary

only if you want ipam-ext to be standalone, there should be kustomize which we dont have now, it is always nice to have, but shouldn't block imho

maiqueb: > i would suggest gladly to keep CNAO do what needed, as we already did for deploying ipam on OpenShift we are doing the same for all components, it is easy and straight forward since the values are binary

Easy and straight-forward ... that's debatable. I don't find it easy, nor straight-forward. I actually find it confusing, complex, and I even need to provide the batteries for it ... 

I don't know what you mean w/ the values are binary - AFAIU @ormergi needs to customize the namespace, and a set of labels. 

> 
> only if you want ipam-ext to be standalone, there should be kustomize which we dont have now, it is always nice to have, but shouldn't block imho

We're relying heavily in kustomize on this repo - it is CNAO that is not aware of kustomize and relies on a bunch of scripts we have to maintain to do a specialized tool's job.

Still, you're the maintainer: I am just an unhappy user representing another unhappy user.

hack/install-deny-all-net-pol.sh
maiqueb, line 1: do we even need this file ? Like, once you have policies in place targetting the component (i.e. the pod), that's the only traffic that gets accepted, isn't it ?

config/networkpolicy/allow-egress-dns.yaml
oshoval: Binary means it has set of fixed values, one for upstream and one for OpenShift,
I appreciate your flexibility.

If we ever desire to change it, we can discuss how and once there is priority,
(imho, it is good as it is now wrt CNAO fwiw).

The only thing that might be nice to have in the future, is that each repo is standalone both on k8s and OpenShift, but there is no rush for it atm, and it can be done orthogonal (even here we didn't do it yet, because there are more important things, which is another reason to not go into unrequired adventures ad hoc, but to plan and prioritize them)

hack/install-deny-all-net-pol.sh
```diff
@@ -0,0 +1,37 @@
...
+readonly ns="$(${KUBECTL} get pod -l app=ipam-virt-workloads -A -o=custom-columns=NS:.metadata.namespace --no-headers | head -1)"
+[[ -z "${ns}" ]] && echo "FATAL: kubevirt-ipam-controller pods not found. Make sure kubevirt-ipam-controller is installed" && exit 1
+
+readonly np_name="deny-all"
+${KUBECTL} -n "${ns}" get networkpolicy "${np_name}" &> /dev/null ||
+  cat <<EOF | ${KUBECTL} -n "${ns}" apply -f -
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: ${np_name}
```
maiqueb, line 29: shouldn't you limit this to your namespace ?

ormergi: Please note that the create call specify the target namespace (${ns}) and ensure the policy is created in ipam-ext namespace.

maiqueb (approved):
/approve
/lgtm

PR #103, opened 2025-05-22
claims: Format key to domain name

qinqon:
Format new ipamclaims keys to follow kubernetes CRD name restrictions

Fixes #https://issues.redhat.com/browse/CNV-62286

There is no problems from upgrades since already created ipamclaims have a proper name that do not need formatting.

**Release note**:

```release-note
claims: Format key to domain name
```

### DIFF
--- a/pkg/claims/claims.go
+++ b/pkg/claims/claims.go
@@ -3,6 +3,8 @@ package claims
 import (
 	"context"
 	"fmt"
+	"regexp"
+	"strings"
 
 	apitypes "k8s.io/apimachinery/pkg/types"
 
@@ -14,7 +16,16 @@ import (
 	virtv1 "kubevirt.io/api/core/v1"
 )
 
-const KubevirtVMFinalizer = "kubevirt.io/persistent-ipam"
+const (
+	KubevirtVMFinalizer      = "kubevirt.io/persistent-ipam"
+	rfc1123SubdomainsPattern = `[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*`
+	claimKeySeparator        = "."
+)
+
+var (
+	// Define the regex pattern for valid RFC 1123 subdomains
+	rfc1123SubdomainsRegexp = regexp.MustCompile(rfc1123SubdomainsPattern)
+)
 
 func Cleanup(c client.Client, vmiKey apitypes.NamespacedName) error {
 	ipamClaims := &ipamclaimsapi.IPAMClaimList{}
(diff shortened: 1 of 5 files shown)


pkg/claims/claims.go
```diff
@@ -43,5 +54,22 @@ func OwnedByVMLabel(vmiName string) client.MatchingLabels {
...
+}
+
+func convertToCRDName(input string) string {
+	// Convert to lowercase
+	crdName := strings.ToLower(input)
+
+	// Find the longest valid substring that matches the pattern
+	matches := rfc1123SubdomainsRegexp.FindAllString(crdName, -1)
+
+	// Join valid substrings with a hyphen (if there are multiple matches)
```
maiqueb, line 67: with a hyphen ?... that's not true - the separator is a '.'

pkg/claims/claims.go
```diff
@@ -43,5 +54,22 @@ func OwnedByVMLabel(vmiName string) client.MatchingLabels {
...
+	crdName = strings.Join(matches, claimKeySeparator)
+
+	// Truncate to 253 characters if necessary
+	if len(crdName) > 253 {
```
maiqueb, line 71: maybe turn 253 into a const ?

pkg/vminetworkscontroller/vmi_controller_test.go
maiqueb, line 1: wouldn't it be helpful to add a test w/ an invalid interface name to exercise the new name creation ?

maiqueb (changes requested):

pkg/claims/claims.go
oshoval: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names
just took about RFC 1123, in case the rules fit here as well
```
RFC 1123 Label Names 
Some resource types require their names to follow the DNS label standard as defined in [RFC 1123](https://tools.ietf.org/html/rfc1123). This means the name must:

contain at most 63 characters
contain only lowercase alphanumeric characters or '-'
start with an alphanumeric character
end with an alphanumeric character
```

aint it should be 63 ?
(maybe any other rule interesting here)

maiqueb: tks for actually reading through the standard Or.

I think @oshoval 's right.

qinqon: I see 253 here -> https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names

pkg/vminetworkscontroller/vmi_controller_test.go
qinqon: make sense, I only added small unit test.

pkg/claims/claims.go
maiqueb: hm, so:
- DNS Subdomain Names: max len => 253
- RFC 1123 Label Names: max len => 63

Seems we got to figure out exactly which we're handling. I think we follow DNS subdomain names - from the [custom resources docs](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#customresourcedefinitions):
>  the name of an object whose kind/resource is defined by a CRD must also be a valid DNS subdomain name

Let's keep 253. Or ... better yet. Check it out. I'm nearly sure you can test this with the fakeclient structs.

qinqon: We have to follow what kubernetes mandates, since the implementation of kubernetes is doing so.

maiqueb (approved):
/approve
/lgtm
/hold

Please remove the hold once you're happy with the PR. 

Thanks for this quick fix !

PR #105, opened 2025-06-19
udn, static ips, webhook: use multus default network annotation to specify MAC and IP addresses for KubeVirt VMs

maiqueb:
This PR changes the mutating webhook implementation to add the multus default network annotation to launcher pods whose VMs have MAC and IP address requests.

Unit tests, and e2e tests are added to assert the correct behavior of this functionality. 

Follow-up PRs which are still required:
- deprecate the currently used `"k8s.ovn.org/primary-udn-ipamclaim"` annotation.
- customize the namespace where the cluster default network NAD is provisioned

Fixes #

**Release note**:

```release-note
Use multus default network annotation to specify MAC and IP addresses for KubeVirt VMs
```


pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -233,54 +276,94 @@ func ensureIPAMClaimRefAtNetworkSelectionElements(ctx context.Context,
...
 		if network.Pod != nil {
-			return network.Name
+			return &network
 		}
 	}
-	return ""
+	return nil
+}
+
+func vmiPodPrimaryNetworkInterface(vmi *virtv1.VirtualMachineInstance, network virtv1.Network) *virtv1.Interface {
```
qinqon, line 336: name it by what it really does 
```suggestion
func vmiNetworkInterface(vmi *virtv1.VirtualMachineInstance, network virtv1.Network) *virtv1.Interface {
```

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -233,54 +276,94 @@ func ensureIPAMClaimRefAtNetworkSelectionElements(ctx context.Context,
...
+}
+
+func primaryNetworkIPRequests(vmi *virtv1.VirtualMachineInstance, podNetworkName string) []string {
+	var addrs map[string][]string
+	ipRequests := vmi.Annotations[config.IPRequestsAnnotation]
+	if err := json.Unmarshal([]byte(ipRequests), &addrs); err != nil {
+		return nil
+	}
+
+	return addrs[podNetworkName]
```
qinqon, line 368: guard agains ipRequest == "" <- there is no ip request annotation

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -233,54 +276,94 @@ func ensureIPAMClaimRefAtNetworkSelectionElements(ctx context.Context,
...
+	return &v1.NetworkSelectionElement{
+		Name:               defaultNetworkName,
+		Namespace:          namespace,
+		IPRequest:          ips,
+		MacRequest:         mac,
+		IPAMClaimReference: ipamClaimName,
+	}
+}
+
+func primaryNetworkIPRequests(vmi *virtv1.VirtualMachineInstance, podNetworkName string) []string {
```
qinqon, line 361: We should name things for what it does, the "primary" part is what the caller is using it for.
```suggestion
func vmiInterfaceIPRequests(vmi *virtv1.VirtualMachineInstance, vmiInterfaceName string) []string {
```

qinqon (changes requested):

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -100,16 +113,41 @@ func (a *IPAMClaimsValet) Handle(ctx context.Context, request admission.Request)
...
 	}
-	if newPrimaryNetworkIPAMClaimName != "" {
+
+	if primaryUDNInterface != nil {
 		if newPod == nil {
 			newPod = pod.DeepCopy()
 		}
-		updatePodWithOVNPrimaryNetworkIPAMClaimAnnotation(newPod, newPrimaryNetworkIPAMClaimName)
+
+		primaryUDNIPRequests := primaryNetworkIPRequests(vmi, primaryUDNInterface.Name)
```
RamLavi: Do you want to return if the Multus Default Network Annotation already lives, or do you want to calculate it every time?

.github/workflows/checks.yaml
```diff
@@ -72,6 +72,8 @@ jobs:
       run: make cluster-sync
 
     - name: Run e2e tests
+      env:
+        E2E_TEST_ARGS: --procs=1
```
RamLavi, line 88: you can just remove the -p flag [here](https://github.com/kubevirt/ipam-extensions/blob/main/Makefile#L86)

Also, can you explain more why they need to run serially? which resources are constrained, and where?

test/e2e/persistentips_test.go
```diff
@@ -216,6 +220,63 @@ var _ = DescribeTableSubtree("Persistent IPs", func(params testParams) {
 			})
 		})
 
+		When("the VM is created with a user defined MAC and IP addresses", func() {
```
RamLavi, line 223: I understand the desire to "add the new feature" to the existing tree, however it makes us use skips that introduce hidden assumptions to test runners.
So let's back up a bit. This looks to me like a [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle) violation — the two VM NIC types are now treated differently by ipam-ext, so the tests are changing for different reasons. Time to split.

Please consider moving this feature, that is only supported on primaryUDN, to a separate DescribeTable:
```
var _ = Describe("User-defined MAC/IP support", func() {
	DescribeTable("on supported network roles", func(params testParams) {
		// move only relevant test logic here
		...
	},
		Entry("primary UDN", testParams{
			role:                 rolePrimary,
			networkInterfaceName: primaryLogicalNetworkInterfaceName,
			vmi:                  vmiWithManagedTap,
		}),
	)
})
```

RamLavi (commented):
Hey, I've added a few comments

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -358,12 +405,56 @@ func multusDefaultNetworkAnnotation(
...
+
+	for _, ip := range primaryNetIPs {
+		var targetSubnet string
+
+		if ips.IsIPv4(ip) {
+			// Find first available IPv4 subnet
+			if len(ipv4Subnets) == 0 {
+				return nil, fmt.Errorf("no IPv4 subnet configured for IPv4 IP request: %s", ip)
+			}
+			targetSubnet = ipv4Subnets[0]
```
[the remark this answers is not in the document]
RamLavi: checked. seems to work
```
func primaryNetworkIPRequests(
	vmi *virtv1.VirtualMachineInstance,
	podNetworkName string,
	primaryNet *config.RelevantConfig,
) ([]string, error) {
	var addrs map[string][]string
	ipRequests, hasIPRequests := vmi.Annotations[config.IPRequestsAnnotation]
	if !hasIPRequests {
		return nil, nil
	}
	if err := json.Unmarshal([]byte(ipRequests), &addrs); err != nil {
		return nil, err
	}

	primaryNetIPs := addrs[podNetworkName]
	result := make([]string, 0, len(primaryNetIPs))

	for _, ipStr := range primaryNetIPs {
		ip := net.ParseIP(ipStr)
		if ip == nil {
			return nil, fmt.Errorf("invalid IP address: %s", ipStr)
		}

		found := false
		subnetList := strings.Split(primaryNet.Subnets, ",")
		for _, subnet := range subnetList {
			subnet = strings.TrimSpace(subnet)
			if subnet == "" {
				continue
			}

			_, ipNet, err := net.ParseCIDR(subnet)
			if err != nil {
				// Log or skip malformed subnet instead of returning
				continue
			}
			if ipNet.Contains(ip) {
				maskSize, _ := ipNet.Mask.Size()
				result = append(result, fmt.Sprintf("%s/%d", ipStr, maskSize))
				found = true
				break
			}
		}

		if !found {
			return nil, fmt.Errorf("no matching subnet found for IP: %s", ipStr)
		}
	}

	return result, nil
}
```

maiqueb (commented):
Replying to comments.

.github/workflows/checks.yaml
maiqueb: Removed this commit. Resolving this comment.

pkg/ipamclaimswebhook/podmutator.go
qinqon: Now I see it covered

pkg/ips/primarynetwork_test.go
```diff
@@ -0,0 +1,231 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			Expect(result).To(BeNil())
+		})
+	})
+
+	When("VMI has IP requests annotation", func() {
+		Context("with valid IPv4 and IPv6 addresses", func() {
+			BeforeEach(func() {
+				ipRequests := map[string][]string{
+					"podnet": {"192.168.1.10", "fd20:1234::200"},
```
RamLavi, line 49: notice that "fd20:1234::200" is not in "fd12:1234::/64" subnet that is set [here](https://github.com/kubevirt/ipam-extensions/pull/105/files#diff-d3e654871f4829cf4633471bb57cb0c485f0839164a9231e9f24bb79ebd4d721R33). 
The configuration seems invalid.
Question is - should we just be a vessel that copies the kubevirt annotation to multus annotation and let OVNK reject this?

pkg/config/types.go
```diff
@@ -8,7 +8,9 @@ import (
 type NetworkRole string
 
 const (
-	NetworkRolePrimary NetworkRole = "primary"
+	NetworkRolePrimary         NetworkRole = "primary"
+	IPRequestsAnnotation       string      = "network.kubevirt.io/addresses"
```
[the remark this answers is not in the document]
maiqueb: I encourage you to read this answer in the enhancement: https://github.com/openshift/enhancements/pull/1793#discussion_r2191603398

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -233,54 +302,117 @@ func ensureIPAMClaimRefAtNetworkSelectionElements(ctx context.Context,
...
+	pluginConfig *config.RelevantConfig,
+) (*virtv1.Interface, error) {
 	log := logf.FromContext(ctx)
-	if pod.Annotations[config.OVNPrimaryNetworkIPAMClaimAnnotation] != "" {
-		return "", nil
+
+	// TODO: once OVN-K lands this feature, we can remove uncomment this code
+	//if !pluginConfig.AllowPersistentIPs {
+	//	return nil, nil
+	//}
```
maiqueb, line 315: @qinqon please remind me: you're already removing this in your https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5321 PR, aren't you ?

qinqon: I bumped nad-client to remove limitation of iprequest + ipamclaimref

pkg/config/types.go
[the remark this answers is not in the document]
RamLavi: I'm not going to insist on this, so take this only as an opinionated suggestion, but stating the annotation name to the target instead of the entity handling it is wrong, and limits it's use.

pkg/ips/primarynetwork_test.go
RamLavi: In terms of visibility I agree the error must be seen by the user where they expect it (i.e. the pod and ipamClaim).
You can resolve this.

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -233,54 +302,117 @@ func ensureIPAMClaimRefAtNetworkSelectionElements(ctx context.Context,
...
+func findPrimaryUDNLogicalNetworkAttachment(
+	ctx context.Context,
+	vmi *virtv1.VirtualMachineInstance,
+	pluginConfig *config.RelevantConfig,
+) (*virtv1.Interface, error) {
 	log := logf.FromContext(ctx)
-	if pod.Annotations[config.OVNPrimaryNetworkIPAMClaimAnnotation] != "" {
-		return "", nil
+
+	// TODO: once OVN-K lands this feature, we can remove uncomment this code
```
qinqon, line 312: Bumping the nad-client is enough we don't need this anymore

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -233,54 +302,117 @@ func ensureIPAMClaimRefAtNetworkSelectionElements(ctx context.Context,
...
+	//if !pluginConfig.AllowPersistentIPs {
+	//	return nil, nil
+	//}
+
+	podNetwork := vmiPodNetwork(vmi)
```
qinqon, line 317: Don't we also want to check there is the "l2bridge" binding ? like we can end up returning a masquerade pod interface and that's wrong.

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -233,54 +302,117 @@ func ensureIPAMClaimRefAtNetworkSelectionElements(ctx context.Context,
 	return hasChangedNetworkSelectionElements, nil
 }
 
-func findNewPrimaryNetworkIPAMClaimName(ctx context.Context,
-	cli client.Client, pod *corev1.Pod, vmName string) (string, error) {
+func findPrimaryUDNLogicalNetworkAttachment(
```
qinqon, line 305:
```suggestion
func findPrimaryUDNInterface(
```

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -233,54 +302,117 @@ func ensureIPAMClaimRefAtNetworkSelectionElements(ctx context.Context,
...
+	ctx context.Context,
+	vmi *virtv1.VirtualMachineInstance,
+	pluginConfig *config.RelevantConfig,
```
qinqon, line 308: We no longer need the pluginConfig

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -233,54 +302,117 @@ func ensureIPAMClaimRefAtNetworkSelectionElements(ctx context.Context,
...
-
-	networkName := vmiPodNetworkName(vmi)
-	if networkName == "" {
-		log.Info("vmi has no pod network primary UDN ipam claim will not be created", "vmi", vmKey.String())
-		return "", nil
-	}
-
-	return claims.ComposeKey(vmi.Name, networkName), nil
+	return vmi, nil
 }
```
qinqon, line 383: We don't need this functions is just retrieving the VMI by name now.

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -233,54 +302,117 @@ func ensureIPAMClaimRefAtNetworkSelectionElements(ctx context.Context,
...
+	return nil
+}
+
+func multusDefaultNetworkAnnotation(
+	namespace string,
+	mac string,
+	ipamClaimName string,
+	ips ...string,
+) *v1.NetworkSelectionElement {
+	const defaultNetworkName = "default"
```
qinqon, line 410: Let's put this constant next to the default namespace one.

pkg/ips/primarynetwork.go
```diff
@@ -0,0 +1,62 @@
...
+		}
+
+		// Extract netmask from subnet
+		splitSubnet := strings.Split(targetSubnet, "/")
+		if len(splitSubnet) != 2 {
+			return nil, fmt.Errorf("invalid subnet format: %s", targetSubnet)
+		}
+		subnetNetmask := splitSubnet[1]
+		result = append(result, fmt.Sprintf("%s/%s", ip, subnetNetmask))
+	}
```
qinqon, line 59: I don't remember do we really need to add the cidr length ? I remember I saw multus code understanding both.

qinqon (changes requested):
Not a biggy

pkg/ipamclaimswebhook/podmutator.go
maiqueb: Good point.

pkg/ipamclaimswebhook/podmutator.go
maiqueb: What do you mean we don't need it ? We return if the plugin doesn't support persistent ips.

pkg/ipamclaimswebhook/podmutator.go
qinqon: fair enough, we can make it more picky at a follow up

pkg/ipamclaimswebhook/podmutator.go
qinqon: > What do you mean we don't need it ? We return if the plugin doesn't support persistent ips.

I see the function just returning the VMI, where is the persisent IPs check ? I am talking about `runningVMForPod` fuction.

qinqon: ahh right, I mixed stuff as always,  so with the udn-client update we no longer need to check persisentips right ? so we don't need pluginConfig, or do we ?

pkg/ipamclaimswebhook/podmutator.go
maiqueb: Done.

pkg/ipamclaimswebhook/podmutator.go
maiqueb: Resolving then; let's make a decision later.

pkg/ips/primarynetwork.go
maiqueb: let's keep it. We can make a decision later on if we can make this simpler. 

Deal ?

pkg/ipamclaimswebhook/podmutator.go
qinqon: Right then at the top of  podmutator.go

qinqon: nah forget let's not stop this for that

qinqon (approved):
LGTM

PR #106, opened 2025-06-26
Fail early if the VMI is using the multus default network feature from the KubeVirt API

RamLavi:
This PR adds a validation that the the vm pod does not have an existing v1.multus-cni.io/default-network pointing to a network which is not the primary-udn one.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/persistentips_test.go
+++ b/test/e2e/persistentips_test.go
@@ -29,6 +29,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -420,6 +421,84 @@ var _ = Describe("Primary User Defined Network attachment", func() {
 		})
 	})
 
+})
+
+var _ = Describe("Webhook validation for default-network annotation", func() {
+	var (
+		failureCount = 0
+		namespace    string
+	)
+
+	JustAfterEach(func() {
+		if CurrentSpecReport().Failed() {
+			failureCount++
+			logFailure(failureCount, namespace)
+		}
+	})
+
+	When("launcher pod has default-network multus annotation pointing to wrong interface", func() {
+		var (
+			td  testenv.TestData
+			vm  *kubevirtv1.VirtualMachine
+			vmi *kubevirtv1.VirtualMachineInstance
+			nad *nadv1.NetworkAttachmentDefinition
+		)
+
+		BeforeEach(func() {
+			td = testenv.GenerateTestData()
+			namespace = td.Namespace
+			td.SetUp(primaryUDNNamespaceLabels())
+			DeferCleanup(func() {
+				td.TearDown()
+			})
+
+			// TODO: delete the code block below once OVN-Kubernetes provisions
+			// the default network NAD
+			const ovnKubernetesNamespace = "ovn-kubernetes"
+			By("Ensuring the cluster default network attachment NetworkAttachmentDefinition exists")
+			Expect(ensureDefaultNetworkAttachmentNAD(ovnKubernetesNamespace)).To(Succeed())
+			// END code to be deleted block
+
+			nad = testenv.GenerateLayer2WithSubnetNAD(nadName, td.Namespace, rolePrimary)
+			By("Create NetworkAttachmentDefinition")
+			Expect(testenv.Client.Create(context.Background(), nad)).To(Succeed())
+
+			vmi = vmiWithManagedTap(td.Namespace)
+			vm = testenv.NewVirtualMachine(vmi, testenv.WithRunning())
+
+			By("Adding wrong default-network annotation that should cause webhook rejection")
+			if vm.Spec.Template.ObjectMeta.Annotations == nil {
+				vm.Spec.Template.ObjectMeta.Annotations = make(map[string]string)
+			}
+			vm.Spec.Template.ObjectMeta.Annotations["v1.multus-cni.io/default-network"] = `[{"name":"wrong_interface"}]`
+		})
+
+		It("should reject VM Pod creation", func() {
+			By("Creating VM with wrong default-network annotation")
+			err := testenv.Client.Create(context.Background(), vm)
+			Expect(err).NotTo(HaveOccurred(), "VM creation should succeed")
+
+			By("Waiting for webhook rejection event to appear")
+			Eventually(func(g Gomega) []corev1.Event {
+				eventsList := corev1.EventList{}
+				err := testenv.Client.List(context.Background(), &eventsList, &client.ListOptions{
+					Namespace: vm.Namespace,
+				})
+				g.Expect(err).NotTo(HaveOccurred())
+				return eventsList.Items
+			}).WithPolling(time.Second).
+				WithTimeout(30*time.Second).
+				Should(ContainElement(And(
+					HaveField("Type", "Warning"),
+					HaveField("Reason", "FailedCreate"),
+					HaveField("InvolvedObject.Kind", "VirtualMachineInstance"),
+					HaveField("Message", ContainSubstring(`multus default network is only allowed on the primary UDN interface "pod_iface", but was requested on interface "wrong_interface"`)),
+				)), "Should find webhook rejection event for VMI pod creation failure")
+
+			By("Verifying the VM has not started due to pod creation failure")
+			Expect(testenv.ThisVMReadiness(vm)()).To(BeFalse(), "VM should not be ready due to webhook rejection")
+		})
+	})
 })
 
 func foregroundDeleteOptions() *client.DeleteOptions {


RamLavi:
@maiqueb PTAL when you can

maiqueb (commented):
I always expected you to actually check the VMI's spec and fail if you found a multus network w/ the `default` knob on.

I was initially surprised about the option you chose, but then realized that it has the advantage of only requiring pod information for it. But, we're already reading the VMI in this function, a little bit further up. Don't you think it is more explicit if we did that instead ? 

I think the approach you chose is OK FWIW.

RamLavi:
> I always expected you to actually check the VMI's spec and fail if you found a multus network w/ the `default` knob on.
> 
> I was initially surprised about the option you chose, but then realized that it has the advantage of only requiring pod information for it. But, we're already reading the VMI in this function, a little bit further up. Don't you think it is more explicit if we did that instead ?
> 
> I think the approach you chose is OK FWIW.

hmm, not sure what you mean by "more explicit". 
I checked the pod because that's the only place that really matters. 
If kubevirt/kubevirt introduces a bug/feature where the VM doesn't copy the annotation for some reason - we want the test to show it and break.

RamLavi:
> > > I always expected you to actually check the VMI's spec and fail if you found a multus network w/ the `default` knob on.
> > > I was initially surprised about the option you chose, but then realized that it has the advantage of only requiring pod information for it. But, we're already reading the VMI in this function, a little bit further up. Don't you think it is more explicit if we did that instead ?
> > > I think the approach you chose is OK FWIW.
> > 
> > hmm, not sure what you mean by "more explicit".
> 
> I mean that it's easier for the reader to understand that you can't use multus default networks on the vms and primary UDNs at the same time if we check that from the API, rather than checking the "how" is that done - i.e. an annotation was already there. 
> 
> > I checked the pod because that's the only place that really matters. If kubevirt/kubevirt introduces a bug/feature where the VM doesn't copy the annotation for some reason - we want the test to show it and break.
> 
> Why would we (ipam-extensions) care about kubevirt/kubevirt breaking *their* default network feature ? It isn't a feature we provide, or even care about. 
> 

We shouldn't care at all, we should only care of the final product. The annotation should end up at the pod, this why I think we should check there. 

It doesn't matter if the annotation was introduced at vm, vmi, or pod, as long as it end up in the pod. 
This is at least my reasoning. 

>

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -190,6 +194,23 @@ func (a *IPAMClaimsValet) Handle(ctx context.Context, request admission.Request)
 	return admission.Allowed("carry on")
 }
 
+func validateDefaultMultusNetworkRequest(pod *corev1.Pod, primaryUDNInterfaceName string) error {
+	defaultNet, err := udn.GetK8sPodDefaultNetworkSelection(
+		pod.Annotations[config.MultusDefaultNetAnnotation],
```
maiqueb, line 199: let's check in the caller is the annotation is actually present instead of passing an empty string down.

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -190,6 +194,23 @@ func (a *IPAMClaimsValet) Handle(ctx context.Context, request admission.Request)
...
+		pod.Namespace)
+	if err != nil {
+		return fmt.Errorf("failed to parse default network annotation: %v", err)
+	}
+
+	if defaultNet != nil && defaultNet.Name != primaryUDNInterfaceName {
+		return fmt.Errorf(
+			"multus default network is only allowed on the primary UDN interface (%s), but was requested on interface %s",
```
maiqueb, line 207: %q maybe ?

pkg/ipamclaimswebhook/podmutator_test.go
```diff
@@ -325,6 +325,24 @@ var _ = Describe("KubeVirt IPAM launcher pod mutato machine", Serial, func() {
 				},
 			},
 		}),
+		Entry("launcher pod with existing default-network multus annotation pointing to the non-udn interface", testConfig{
```
maiqueb, line 328: can you ensure we have a unit test for a primary UDN launcher pod that is accepted ?

pkg/udn/default_network_selection.go
```diff
@@ -0,0 +1,29 @@
+package udn
+
+import (
+	"fmt"
+
+	v1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+	nadutils "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/utils"
+)
+
+func GetK8sPodDefaultNetworkSelection(
```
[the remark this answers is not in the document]
maiqueb: let's re-write in whatever fashion that makes sense to us - if needed.

Trying to say I don't think you have to import anything.

maiqueb:
@RamLavi did you figure out why the e2e tests were failing ? let me know if you need help.

pkg/ipamclaimswebhook/podmutator_test.go
RamLavi: we have of course, but let's add another one to check accepted case where the default-multus annotation is already there (from prior reconciles)

RamLavi:
> @RamLavi did you figure out why the e2e tests were failing ? let me know if you need help.

well at the time the test ran - not all the ovnk code bits were in. we'll get a clearer picture after I push :)

RamLavi:
[Change](https://github.com/kubevirt/ipam-extensions/compare/f05b81c6c13c16aac420e18eb326adeb5492d73a..1a556bf1d4f9976dfc53d9f3bd182dff8a92c751): fixes to @maiqueb 's review (exepct e2e - waiting for @maiqueb call on this)

maiqueb:
I still think this change merits an e2e test.

RamLavi:
[Change](https://github.com/kubevirt/ipam-extensions/compare/1a556bf1d4f9976dfc53d9f3bd182dff8a92c751..b04645d1485673735b24f6e32185289341230213): fix linter

maiqueb:
@RamLavi the unit test fails w/ 
```
      <v1.AdmissionResponse>: {
          UID: "",
          Allowed: false,
          Result: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "",
              Message: "multus default network is only allowed on the primary UDN interface (\"podnet\"), but was requested on interface \"not-podnet\"",
              Reason: "Forbidden",
              Details: nil,
              Code: 403,
          },
          Patch: nil,
          PatchType: nil,
          AuditAnnotations: nil,
          Warnings: nil,
      }
  to equal
      <v1.AdmissionResponse>: {
          UID: "",
          Allowed: false,
          Result: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "",
              Message: "multus default network is only allowed on the primary UDN interface (podnet), but was requested on interface not-podnet",
              Reason: "Forbidden",
              Details: nil,
              Code: 403,
          },
          Patch: nil,
          PatchType: nil,
          AuditAnnotations: nil,
          Warnings: nil,
      }
```

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -190,6 +194,26 @@ func (a *IPAMClaimsValet) Handle(ctx context.Context, request admission.Request)
...
+	}
+
+	defaultNet, err := udn.GetK8sPodDefaultNetworkSelection(annotationValue, pod.Namespace)
+	if err != nil {
+		return fmt.Errorf("failed to parse default network annotation: %v", err)
+	}
+
+	if defaultNet != nil && defaultNet.Name != primaryUDNInterfaceName {
+		return fmt.Errorf(
+			"multus default network is only allowed on the primary UDN interface (%q), but was requested on interface %q",
```
maiqueb, line 210:
```suggestion
			"multus default network is only allowed on the primary UDN interface %q, but was requested on interface %q",
```

no point having the parenthesis now that you've quoted the variable.

pkg/ipamclaimswebhook/podmutator_test.go
```diff
@@ -325,6 +325,51 @@ var _ = Describe("KubeVirt IPAM launcher pod mutato machine", Serial, func() {
...
+			inputNADs: []*nadv1.NetworkAttachmentDefinition{
+				dummyPrimaryNetworkNAD(nadName),
+			},
+			inputPod: dummyPodForVMWithAnnotation("" /*without network selection element*/, vmName,
+				map[string]string{config.MultusDefaultNetAnnotation: "[{\"name\":\"not-podnet\"}]"}),
+			expectedAdmissionResponse: admissionv1.AdmissionResponse{
+				Allowed: false,
+				Result: &metav1.Status{
+					Message: "multus default network is only allowed on the primary UDN interface (podnet), " +
+						"but was requested on interface not-podnet",
```
maiqueb, line 340: you're missing quotes around `podnet` and `not-podnet` .

```suggestion
					Message: "multus default network is only allowed on the primary UDN interface \"podnet\", " +
						"but was requested on interface \"not-podnet\"",
```

maiqueb (changes requested):

pkg/ipamclaimswebhook/podmutator_test.go
RamLavi: DONE

RamLavi:
[Change](https://github.com/kubevirt/ipam-extensions/compare/97928e124ad0271ed5564791c57f6017b2b11166..044365ee79c72925aee5c8a1bb66f4fe6d52dc72): Remove parenthesis from error message

RamLavi:
@maiqueb I will run this locally and figure out what's wrong with the tests panic

RamLavi:
[Change](https://github.com/kubevirt/ipam-extensions/compare/471bfd677b464d35a86cffcb43258158f14a1288..a47e2a0abef1fdd783c3e4bb844204e580fd4d48): fix issue in e2e added

RamLavi:
[Change](https://github.com/kubevirt/ipam-extensions/compare/a47e2a0abef1fdd783c3e4bb844204e580fd4d48..609b3c7e34ec0532630601b953c18315ad780e60): Fix e2e test

PR #110, opened 2025-07-07
feat: use branch name for image tagging while keeping :latest for backward compatibility

qinqon:
- Update GitHub Actions workflow to publish images with both branch-based tags and :latest
- Add branch name extraction and sanitization for Docker tag compatibility
- Template manifests using branch-based tags for branch pushes
- Keep :latest tagging for backward compatibility with existing manifests
- Update install manifest to reflect current configuration

This enables better image tracking per branch while maintaining compatibility with existing deployments using :latest tags.

**Release note**:

```release-note
Push to branch name when merged PRs.
```

### DIFF
--- a/.github/workflows/publish-img.yaml
+++ b/.github/workflows/publish-img.yaml
@@ -29,6 +29,20 @@ jobs:
         run: |
           echo "git_commit_hash=$(git describe --no-match  --always --abbrev=8 --dirty)" >> $GITHUB_ENV
 
+      - name: Extract branch name
+        shell: bash
+        run: |
+          # Extract branch name and sanitize it for Docker tag compatibility
+          if [[ "${{ github.ref }}" == "refs/heads/"* ]]; then
+            BRANCH_NAME="${{ github.ref#refs/heads/ }}"
+            # Replace invalid characters with dashes and convert to lowercase
+            SANITIZED_BRANCH=$(echo "$BRANCH_NAME" | sed 's/[^a-zA-Z0-9._-]/-/g' | tr '[:upper:]' '[:lower:]')
+            echo "branch_tag=${SANITIZED_BRANCH}" >> $GITHUB_ENV
+            echo "is_branch_push=true" >> $GITHUB_ENV
+          else
+            echo "is_branch_push=false" >> $GITHUB_ENV
+          fi
+
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3.0.0
 
@@ -40,13 +54,15 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Push latest container image
-        if: github.repository_owner == 'kubevirt'
+      - name: Push branch-based and latest container images
+        if: github.repository_owner == 'kubevirt' && env.is_branch_push == 'true'
         uses: docker/build-push-action@v5.3.0
         with:
           context: .
           push: true
-          tags: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:latest
+          tags: |
+            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:latest
+            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ env.branch_tag }}
           file: Dockerfile
           platforms: ${{ env.BUILD_PLATFORMS }}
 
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-img.yaml
```diff
@@ -29,6 +29,20 @@ jobs:
         run: |
           echo "git_commit_hash=$(git describe --no-match  --always --abbrev=8 --dirty)" >> $GITHUB_ENV
 
+      - name: Extract branch name
+        shell: bash
+        run: |
+          # Extract branch name and sanitize it for Docker tag compatibility
+          if [[ "${{ github.ref }}" == "refs/heads/"* ]]; then
+            BRANCH_NAME="${{ github.ref#refs/heads/ }}"
```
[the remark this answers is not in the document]
oshoval: combination of tag in a branch might need extra condition ?
`The type of ref that triggered the workflow run. Valid values are branch or tag.`

.github/workflows/publish-img.yaml
```diff
@@ -40,13 +51,15 @@ jobs:
...
-        if: github.repository_owner == 'kubevirt'
+      - name: Push branch-based and latest container images
+        if: github.repository_owner == 'kubevirt' && github.ref_type == 'branch'
         uses: docker/build-push-action@v5.3.0
         with:
           context: .
           push: true
-          tags: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:latest
+          tags: |
+            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:latest
```
maiqueb, line 61: wait, iiuc - I might be wrong - this means that pushing to a branch will also overwrite the latest img tag. We don't want that.

We only want to update the `latest` img tag on pushes to the "main" branch.

.github/workflows/publish-img.yaml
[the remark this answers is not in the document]
qinqon: There is the "- name: Push stable container image" already present for that.

.github/workflows/publish-img.yaml
maiqueb: No, I prefer to see everything at once, otherwise I don't understand what's happening. 

This file is for the main branch *or* for tags named `vX.Y.Z` . You're adding a condition which I think skips the tags, hence this will only apply for the main branch. 

Thus, I don't understand what you're fixing in this PR. The main branch is kind of OK, right ? We're simply missing configurations for the other branches.

Makefile
```diff
@@ -1,6 +1,7 @@
 
 # Image URL to use all building/pushing image targets
-IMG ?= kubevirt-ipam-controller:latest
+# Default to registry path for consistency with CI/CD
+IMG ?= ghcr.io/kubevirt/ipam-controller:latest
```
maiqueb, line 4: maybe we want to be consistent w/ the other branches and consume the branch based flavor ?

.github/workflows/checks.yaml
```diff
@@ -29,6 +29,18 @@ jobs:
...
+    - name: Extract branch name
+      shell: bash
+      run: |
+        # Extract branch name and sanitize it for Docker tag compatibility
+        if [[ "${{ github.ref_type }}" == "branch" ]]; then
+          BRANCH_NAME="${{ github.ref_name }}"
+          # Replace invalid characters with dashes and convert to lowercase
+          SANITIZED_BRANCH=$(echo "$BRANCH_NAME" | sed 's/[^a-zA-Z0-9._-]/-/g' | tr '[:upper:]' '[:lower:]')
+          echo "branch_tag=${SANITIZED_BRANCH}" >> $GITHUB_ENV
+        fi
```
maiqueb, line 41: why do we need this here ? this would only be required on img push workflows right ?

maiqueb (changes requested):
Let's forget about this PR, we will need to remember to cherry-pick the commits to the

.github/workflows/publish-img.yaml
```diff
@@ -40,8 +49,18 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Push branch-based container images
+        if: github.repository_owner == 'kubevirt' && github.ref_type == 'branch'
```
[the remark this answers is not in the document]
qinqon: Right, people may want to have their own fork, in fact removing that we can test merge actions at a fork, let me remove it.

PR #112, opened 2025-07-07
[release-0.2.1] feat: use branch name for image tagging

qinqon:
- Update GitHub Actions workflow to publish images with branch name
- Add branch name extraction and sanitization for Docker tag compatibility
- Template manifests using branch-based tags for branch pushes
- Update install manifest to reflect current configuration

```release-note
NONE
```

### DIFF
--- a/.github/workflows/publish-img.yaml
+++ b/.github/workflows/publish-img.yaml
@@ -30,6 +30,17 @@ jobs:
         run: |
           echo "git_commit_hash=$(git describe --no-match  --always --abbrev=8 --dirty)" >> $GITHUB_ENV
 
+      - name: Extract branch name
+        shell: bash
+        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/publish-img.yaml
```diff
@@ -88,6 +99,10 @@ jobs:
           build-args: |
             KUBEVIRT_VERSION=${{ env.KUBEVIRT_VERSION }}
 
+      - name: Template branch manifests
+        if: github.repository_owner == 'kubevirt' && github.ref_type == 'branch'
+        run: IMG=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ env.branch_tag }} make build-installer
```
maiqueb, line 104: why doesn't this re-use the stanza below ? i.e. template release manifests is already doing this for tags. 
Now it seems you want that also for a specific branch.

PR #129, opened 2025-10-30
e2e: use Eventually for IP assertions to avoid race condition

RamLavi:
In the e2e tests there is a hidden assumption that once the VMI is in Ready=true condition then the status.interface.ipAddress is set. This is not guaranteed, causing a [flake](https://github.com/kubevirt/cluster-network-addons-operator/actions/runs/18906135336/job/53964765741?pr=2440).
This PR adds Eventually with 30s timeout to wait for IPs to appear.

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/test/e2e/persistentips_test.go
+++ b/test/e2e/persistentips_test.go
@@ -104,19 +104,17 @@ var _ = DescribeTableSubtree("Persistent IPs", func(params testParams) {
 				By("Creating VM with secondary attachments")
 				Expect(testenv.Client.Create(context.Background(), vm)).To(Succeed())
 
-				By(fmt.Sprintf("Waiting for readiness at virtual machine %s", vm.Name))
-				Eventually(testenv.ThisVMReadiness(vm)).
-					WithPolling(time.Second).
-					WithTimeout(5 * time.Minute).
-					Should(BeTrue())
-
 				By("Wait for IPAMClaim to get created")
 				Eventually(testenv.IPAMClaimsFromNamespace(vm.Namespace)).
 					WithTimeout(time.Minute).
 					WithPolling(time.Second).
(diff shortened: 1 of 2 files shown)


test/e2e/persistentips_test.go
```diff
@@ -116,7 +116,10 @@ var _ = DescribeTableSubtree("Persistent IPs", func(params testParams) {
 					WithPolling(time.Second).
 					ShouldNot(BeEmpty())
 
-				Expect(testenv.ThisVMI(vmi)()).Should(testenv.MatchIPsAtInterfaceByName(params.networkInterfaceName, Not(BeEmpty())))
+				Eventually(testenv.ThisVMI(vmi)).
+					WithPolling(time.Second).
+					WithTimeout(30 * time.Second).
+					Should(testenv.MatchIPsAtInterfaceByName(params.networkInterfaceName, Not(BeEmpty())))
```
maiqueb: I wonder; while this seems OK, isn't it more conceptually correct to have something like a `ThisVMReadiness` wrapper that also checks if the interfaces have IPs ? Or decorate the expectation somehow ?

maiqueb (commented):
just a thought

maiqueb:
> @maiqueb until this is fixed, the test will keep flaking on CNAO ipam-ext tier1 test ([example](https://github.com/kubevirt/cluster-network-addons-operator/actions/runs/19223438291/job/54945599132?pr=2467)) - making me investigate it every time. Let's free ourselves from this burden. How do you want to proceed with this PR?

I worry that patching the immediate problem with this fix is just introducing tech debt that I worry will never be tackled.

So yes, you will no longer be nagged. But the test just got harder to follow and understand. 

I would prefer if we did this properly.

PR #131, opened 2025-11-05
release-0.2.1: fix(manager) Improve upgrade behaviour

qinqon:
Backport from:
- https://github.com/kubevirt/ipam-extensions/pull/128

**Release note**:

```release-note
Increase manager replicas to 2
```


qinqon:
> Shouldn't the first commit be a cherry-pick w/ `-x` option ? so we can track where the commit came from.

Right let me do that

PR #137, opened 2026-01-14
podmutator: Fetch VMI from cluster API when missing from cache

ormergi:
Recently we see the following warning on VMI get calls from kube-api server, coming from the pod-mutator webhook:
```
Unexpected Warning event received: testvmi-hltph,a09eb155-3049-4646-8838-69e8e3360a71: Error creating pod: admission webhook "ipam-claims.k8s.cni.cncf.io" denied the request: failed to access the VMI running in pod "kubevirt-test-default1/": VirtualMachineInstance.kubevirt.io "testvmi-hltph" not found
```
Currently the VMI is fetched using the controller-runtime client which queries the informer cache.
As with webhook, the subject object is given as part of the request, in this case the virt-launcher pod.
In some edge cases, the controller-runtime client informer cache has not been populated with the pod associated VMI yet, resulting with the above warning.

This PR changes fetching VMI logic to poll against the client informer cache when not found.

Fixes #

 Assisted by Claude (Anthropic AI Assistant)

**Release note**:

```release-note
podmutator: When VMI object not found poll against the client informer cache.
```

### DIFF
--- a/pkg/ipamclaimswebhook/podmutator.go
+++ b/pkg/ipamclaimswebhook/podmutator.go
@@ -127,9 +127,7 @@ func (a *IPAMClaimsValet) Handle(ctx context.Context, request admission.Request)
 		return admission.Errored(http.StatusInternalServerError, err)
 	}
 	if hasChangedNetworkSelectionElements {
-		if newPod == nil {
-			newPod = pod.DeepCopy()
-		}
+		newPod = pod.DeepCopy()
 		if err := updatePodSelectionElements(newPod, networkSelectionElements); err != nil {
 			return admission.Errored(http.StatusInternalServerError, err)
 		}


RamLavi (commented):
Hey @ormergi, I think its a good call to use the APIReader

note that the VMI is being called also on [ensureIPAMClaimRefAtNetworkSelectionElements](https://github.com/kubevirt/ipam-extensions/blob/70a2bfd416c035907beef16b81ae6b3048ab3139/pkg/ipamclaimswebhook/podmutator.go#L313) - Please consider passing the VMI to that function in order to avoid doing multiple API calls ( it fetches the VMI x times the amount of NADs) now that we are directly using the API

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -118,7 +123,7 @@ func (a *IPAMClaimsValet) Handle(ctx context.Context, request admission.Request)
 
 	vmKey := types.NamespacedName{Namespace: pod.Namespace, Name: vmName}
 	vmi := &virtv1.VirtualMachineInstance{}
-	if err := a.Get(context.Background(), vmKey, vmi); err != nil {
+	if err := get(ctx, a.Client, a.apiReader, vmKey, vmi); err != nil {
```
akalenyu: tbh at this point I wonder if a non cached client is the right choice for this webhook.
amplification shouldn't be terrible, and this more or less only goes off on creates (update interfaces not common?)

what are the other implications of this webhook operating on a stale vmi?

ormergi:
> Hey @ormergi, I think its a good call to use the APIReader
> 
> note that the VMI is being called also on [ensureIPAMClaimRefAtNetworkSelectionElements](https://github.com/kubevirt/ipam-extensions/blob/70a2bfd416c035907beef16b81ae6b3048ab3139/pkg/ipamclaimswebhook/podmutator.go#L313) - Please consider passing the VMI to that function in order to avoid doing multiple API calls ( it fetches the VMI x times the amount of NADs) now that we are directly using the API

At the moment we dont use API directly on ensureIPAMClaimRefAtNetworkSelectionElements.
But good call, I will do that on a follow up PR because it require some refactoring, is that alright?

pkg/ipamclaimswebhook/podmutator.go
maiqueb: one think to try is polling from the cache ... I mean ... it'll be there eventually - OR something really wrong happened.

maiqueb:
> > Hey @ormergi, I think its a good call to use the APIReader
> > note that the VMI is being called also on [ensureIPAMClaimRefAtNetworkSelectionElements](https://github.com/kubevirt/ipam-extensions/blob/70a2bfd416c035907beef16b81ae6b3048ab3139/pkg/ipamclaimswebhook/podmutator.go#L313) - Please consider passing the VMI to that function in order to avoid doing multiple API calls ( it fetches the VMI x times the amount of NADs) now that we are directly using the API
> 
> At the moment we dont use API directly on ensureIPAMClaimRefAtNetworkSelectionElements. But good call, I will do that on a follow up PR because it require some refactoring, is that alright?

Let's have it here please. We need to evaluate if reading from the API throughout the code base is something we'd like to do.

pkg/ipamclaimswebhook/podmutator.go
oshoval: > one think to try is polling from the cache ... I mean ... it'll be there eventually - OR something really wrong happened.

why not just ask to ignore that warning, since it is a legit scenario which will eventually reconcile this way or another (i.e either by adding read, but it seems like double mechanism, or by high level reconcile which imho is better)

it reminds the OVN controller async creation which require some grace until pod can be created, which is legit to be eventually consistent

maiqueb:
/hold

Let's re-think the approach for a sec here. 
@qinqon can you take a look here ? Can you think of a safer alternative ?

pkg/ipamclaimswebhook/podmutator.go
oshoval: remove the event, print it just if it happened consistently for 2m or so

ormergi:
> I don't understand why we'd want to get from the API at all. Why wouldn't we want to retry from the cache, which will eventually be synced with the results from the API ?

@maiqueb 
The thing is we cant tell when the cache will be synced and have the missing object, it can take ms or minutes in case of some issue. 
The webhook may end up poling and keep blocking the pipeline for few seconds (whatever timeout will be set) for nothing, before it query the API.

I think the middle ground is to fetching from API once in case object not found in cache.
It guarantee we get the object at the same webhook run and not pressuring the API server.

akalenyu:
> > I think that's still fine. But it only queries the API on a subset of updates, no?
> 
> I am afraid it always fetch the VMI (for VM pods who are connected to primary UDN or secondary networks).

I think it's still okay tbh, as the webhook performs a bound number of API calls on every update. I'm sure you can minimize the "every update" bit too. We could ask @vladikr for some general kubevirt side input on api vs cache in webhooks.

pkg/ipamclaimswebhook/client.go
```diff
@@ -0,0 +1,47 @@
...
+				return false, nil // obj not found, keep trying
+			}
+			return true, nil
+		},
+	)
+	if err != nil {
+		return fmt.Errorf("%v: %w", err, gerr)
+	}
+	return nil
+}
```
qinqon: You have to retrie just in NotFound error, you have this to do exactly that https://pkg.go.dev/k8s.io/client-go/util/retry

This should be a internal get not a general purpose public Get function

so call the function
```golang
getObjectWithRetryOnNotFound or something super explicit.
```

qinqon (changes requested):

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -127,9 +127,7 @@ func (a *IPAMClaimsValet) Handle(ctx context.Context, request admission.Request)
 		return admission.Errored(http.StatusInternalServerError, err)
 	}
 	if hasChangedNetworkSelectionElements {
-		if newPod == nil {
-			newPod = pod.DeepCopy()
-		}
+		newPod = pod.DeepCopy()
```
maiqueb: hm, got it ... maybe, to be more explicit, we should declare + initialize in the same call ?

maiqueb (changes requested):
tks. nearly there.

pkg/ipamclaimswebhook/podmutator.go
```diff
@@ -312,8 +313,7 @@ func ensureIPAMClaimRefAtNetworkSelectionElements(
 			"network", pluginConfig.Name,
 		)
 
-		vmiNets := vmiSecondaryNetworks(vmi)
-		networkName, foundNetworkName := vmiNets[nadKey.String()]
+		networkName, foundNetworkName := vmiSpecNetworks[nadKey.String()]
```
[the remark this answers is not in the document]
ormergi: Before this change the vmiNets was part of that for lop and calcualted on each iteration, this change move it out of the for loop so it calculated once.

pkg/ipamclaimswebhook/client.go
ormergi: > You have to retrie just in NotFound error, you have this to do exactly that https://pkg.go.dev/k8s.io/client-go/util/retry

Done

> This should be a internal get not a general purpose public Get function
so call the function
getObjectWithRetryOnNotFound or something super explicit.

@qinqon 
The same issue could manifest when podmutator fetch other kind of object such as the associated NAD.
The function is generalized to allow podmutator to fetch other kind of objects in the same manner; poll against cache when not found. And exported to encourage usage around podmutator.
With this we can ensure fetching objects by podmutator is done in the same way for all kind of objects.

In follow up PR we can do the same for the NAD, fetch it once and  early in the flow, when not found poll from cache..

WDYT?

qinqon: Also NAD and VMs use case is different, if the NAD do not exist the VM is not going to be even created (and kubevirt is using informers too), this is a race between virt-launcher, VM and controller-runtime cached client.

qinqon: ok, then at least don't call this just "get" call this "getWithRetryOnNotFound" something kind of long and awful so we don't go and use it always (this is specific for webhook for example, since Reconcile will re-reconcile and there is no issue with that, in fact is the correct pattern)

pkg/ipamclaimswebhook/client.go
```diff
@@ -0,0 +1,47 @@
...
+	getCtx, cancel := context.WithTimeout(ctx, defaultClientTimeout)
+	defer cancel()
+	log := logr.FromContextOrDiscard(getCtx)
+	const pollImmediately = true
+	objGVK := outObj.GetObjectKind().GroupVersionKind().String()
+	var gerr error
+	err := wait.PollUntilContextCancel(getCtx, defaultClientInterval, pollImmediately,
+		func(context.Context) (bool, error) {
+			if gerr = cli.Get(ctx, key, outObj); gerr != nil {
+				log.Info("Failed to fetch object with", "key", key,
```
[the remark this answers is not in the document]
ormergi: Done

pkg/ipamclaimswebhook/client.go
```diff
@@ -0,0 +1,51 @@
...
+	ctx context.Context,
+	cli crclient.Client,
+	key crclient.ObjectKey,
+	outObj crclient.Object,
+) error {
+	const pollImmediately = true
+	log := crlog.FromContext(ctx)
+	objGVK := outObj.GetObjectKind().GroupVersionKind().String()
+	var gerr error
+	err := k8swait.PollUntilContextTimeout(ctx, defaultClientInterval, defaultClientTimeout, pollImmediately,
```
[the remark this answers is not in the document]
ormergi: This is so much better! thanks for the ref.
Done

qinqon (approved):
LGTM

PR #149, opened 2026-04-20
tls: Allow insecure cipher suites 

ormergi:
Currently insecure ciphers are not allowed for use.
Cluster using insecure ciphers and old TLS settings will break on updates (e.g.: Openshift cluster using TLSProfile of type Old)

This PR allows using insecure ciphers.

Fixes #

Related issue https://redhat.atlassian.net/browse/CNV-84496

**Release note**:

```release-note
Allow TLS insecure ciphers for backward compatibility.
```


pkg/config/tls.go
```diff
@@ -43,15 +43,12 @@ var tlsCurveIDByName = map[string]tls.CurveID{
...
 	for _, cipherSuite := range tls.CipherSuites() {
 		tlsCipherSuiteIDByName[cipherSuite.Name] = cipherSuite.ID
 	}
-
-	for _, insecureCipherSuite := range tls.InsecureCipherSuites() {
-		indexedInsecureCipherSuiteNames[insecureCipherSuite.Name] = struct{}{}
+	for _, cipherSuite := range tls.InsecureCipherSuites() {
+		tlsCipherSuiteIDByName[cipherSuite.Name] = cipherSuite.ID
 	}
 }
```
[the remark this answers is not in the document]
ormergi: Go common convention is init() take presence over other components, including exported functions.
In addition this PR doesnt change the init() or exported function position.
Such change should be done in follow-up PR.

pkg/config/tls.go
```diff
@@ -43,15 +43,12 @@ var tlsCurveIDByName = map[string]tls.CurveID{
...
 func init() {
 	for _, cipherSuite := range tls.CipherSuites() {
 		tlsCipherSuiteIDByName[cipherSuite.Name] = cipherSuite.ID
 	}
-
-	for _, insecureCipherSuite := range tls.InsecureCipherSuites() {
-		indexedInsecureCipherSuiteNames[insecureCipherSuite.Name] = struct{}{}
+	for _, cipherSuite := range tls.InsecureCipherSuites() {
+		tlsCipherSuiteIDByName[cipherSuite.Name] = cipherSuite.ID
 	}
```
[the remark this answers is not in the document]
ormergi: Done

pkg/config/tls.go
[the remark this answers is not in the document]
maiqueb: We can add that to the agents file.

cmd/main.go
```diff
@@ -114,7 +114,12 @@ please see https://pkg.go.dev/crypto/tls#CurveID`,
 
 	ctrl.SetLogger(klog.NewKlogr())
 
-	flagsTLSOpts, err := config.ParseTLSOptions(tlsMinVersionRaw, tlsCipherSuitesRaw, tlsCurvePreferencesRaw)
+	flagsTLSOpts, err := config.ParseTLSOptions(
+		tlsMinVersionRaw,
+		tlsCipherSuitesRaw,
+		tlsCurvePreferencesRaw,
+		setupLog.Info,
```
maiqueb, line 121: 1. why info ? 
2. can we have this as optional ? maybe either use a builder pattern, or just make this attribute variadic. imho it does not make sense to test this in the unit tests.

maiqueb (commented):
I wonder if we always need to pass the logger around. for instance, from production code, for sure yes. But from the unit tests ?...

pkg/config/tls.go
```diff
@@ -43,22 +43,20 @@ var tlsCurveIDByName = map[string]tls.CurveID{
...
-var indexedInsecureCipherSuiteNames = map[string]struct{}{}
-
 func init() {
 	for _, cipherSuite := range tls.CipherSuites() {
 		tlsCipherSuiteIDByName[cipherSuite.Name] = cipherSuite.ID
 	}
-
-	for _, insecureCipherSuite := range tls.InsecureCipherSuites() {
-		indexedInsecureCipherSuiteNames[insecureCipherSuite.Name] = struct{}{}
+	for _, cipherSuite := range tls.InsecureCipherSuites() {
```
qinqon, line 50: Shouldn't we make this feature opt-in ? so by default we don't support insecure ciphers so users has to really activate it ? and also we log a warning when they activate it ?

qinqon (changes requested):
I think this should be opt-in

pkg/config/tls.go
maiqueb: This is a good point. I agree with the flag - `allowInsecureCyphers` ?... Which would be enabled by default, and we'd disable on the next OpenShift release ? ... 

@qinqon or do you think it should be disabled by default ? point is most customers will break on upgrades to the next openshift release unless we allow them to use insecure ciphers.

qinqon: @maiqueb opt-in means disabled by default, we should not ship a insecure product by default, this has be stated by users same as the TLS verification flag on projects.

maiqueb: So we just let them know in the release notes they need to enable this flag if running insecure ciphers ?

ormergi: The project will not use insecure ciphers by default.
By default the project will use TLS 1.3 and use a secured cipher suites (managed by Go runtime).

One will need to to explicitly set older TLS version and provide the appropriate TLS cipher suites, with the --tls-min-version and --tls-cipher-suites.

Adding another flag for allowing insecure ciphers sound over-kill to me;
one will need to actively configure the project deployment to use insecure cipher suites,
defaults use secure TLS version and cipher suites.

qinqon: and what happend if they do it by mistake, or they are just upgrading from old version where TLS are suppose to be secure at the time ? they will be need to check logs to see the warning ? I would add a flag to be able to configure insecure stuff.

cmd/main.go
ormergi: 1. The project use the controller-runtime logger, which provides either Info() or Error(). I interprete this as warning thus I used Info with WARNING prefix.
2. Usually I would go with it and make this optional.
In this particular case I rather keep it simple as possible and defer generalizing it to when there is at least another usage. Any generalization we will do know may not make sense to the next usage.
In unit-test I can just switch to some dummy STUB.
WDYT?

pkg/config/tls.go
ormergi: When the project TLS settings are miss-configured, wrong cipher, once can change the flag value and restart pods. You right, it wont be visible, but if one mess with custom TLS settings they need to know what they are doing.

This project is yet another building block in Kubevirt ecosystem, in practice it managed by CNAO including TLS settings via the TLSProfile API. The TLSProfile of type Old or Custom, allow using insecure cipher by design. 
Having allow-insecure-tls flag will require the higher layer controller (e.g.: CNAO) to digest the selected TLS profile and set the 'insecure-tls' flag once it detect a cipher suite **considered insecure by Go crypto**.

FWIW we can emit warning event on the controller pods for better visibility.

@maiqueb @qinqon WDYT?

RamLavi (approved):
/lgtm

pkg/config/tls.go
qinqon: fair enough

cmd/main.go
qinqon: Ok, make sense

